PR #8, opened 2024-11-16
created user endpoints

SeabertYuan:


src/main/java/com/danielagapov/spawn/Controllers/TestController.java
```diff
@@ -46,4 +58,43 @@ public void setName(String name) {
...
+    private String username;
+    private String firstName;
+    private String lastName;
+    private String bio;
+    
+    public String getUsername() {
+        return this.username;
+    }
+
+    public String getfirstName() {
```
Daggerpov, line 73:
```suggestion
    public String getFirstName() {
```
Suggestion example: Use proper camel case

Daggerpov (changes requested):
Please change all getters and setters to proper camel case. e.g. `getFirstName()`

PR #23, opened 2024-11-19
GET /events/user/{id}

Shane-M-1:
An events endpoint for getting events by user id


src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -25,6 +27,12 @@ public Event getEvent(@PathVariable Long id) {
         return eventService.getEventById(id);
     }
 
+    // full path: /api/v1/events/user/{id}
+    @GetMapping("events/{userID}")
```
Daggerpov, line 31: This mapping should match your full path; I think you might've made a typo

```suggestion
    @GetMapping("events/user/{userID}")
```

Daggerpov (changes requested):
Just one typo, but aside from that, great work!

Daggerpov (approved):
Great work!

PR #29, opened 2024-11-20
added junction table between two users and a friend tag id, based on …

Daggerpov:
Implemented the junction table as we discussed, with the two user ids and a friend tag id

image


Daggerpov:
Revised version, after removing (possibly) unnecessary `UserFriendTag` table:

image

evannawfal (approved):
We are leaving UserFriendTag for easier/cleaner implementation.

PR #31, opened 2024-11-20
PUT /events/{id}, GET /events/friendTag{id}

Shane-M-1:
endpoints for PUT event by id and GET events by friendTag


SeabertYuan (changes requested):
GET endpoints look good but I think I've done a more thorough PUT endpoints in PR #32

Shane-M-1:
> more thorough PUT endpoints in PR #32

Agreed. I removed my PUT endpoint

SeabertYuan (changes requested):
Make sure that the updateEventById method is also removed from IEventService

SeabertYuan (approved):
Looks good to me!

PR #32, opened 2024-11-20
updated models with mutable classes, added appropriate getters and se…

SeabertYuan:
…tters, added PUT methods in controllers

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -44,4 +44,10 @@ public class EventController {
     public Event createEvent(@RequestBody Event newEvent) {
         return eventService.saveEvent(newEvent);
     }
+
+    // full path: /api/v1/events/{id}
+    @PutMapping("{id}")
+    public Event replaceEvent(@RequestBody Event newEvent, @PathVariable Long id) {
+        return eventService.replaceEvent(newEvent, id);
+    }
 }
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -26,11 +26,17 @@ public class EventController {
     public Event getEvent(@PathVariable Long id) {
         return eventService.getEventById(id);
     }
-
+    
     // full path: /api/v1/events/user/{id}
-    @GetMapping("user/{userID}")
-    public List<Event> getEventsByUserID(@PathVariable Long userID) {
-        return eventService.getEventsByUserId(userID);
+    @GetMapping("user/{userId}")
+    public List<Event> getEventsByUserId(@PathVariable Long userId) {
+        return eventService.getEventsByUserId(userId);
+    }
+
+    // full path: /api/v1/events/friendTag/{tagId}
+    @GetMapping("friendTag/{tagId}")
+    public List<Event> getEventsByFriendTagId(@PathVariable Long tagId) {
+        return eventService.getEventsByTagId(tagId);
     }
 
     // full path: /api/v1/events/mock-endpoint
--- a/src/main/java/com/danielagapov/spawn/Models/ChatMessage/ChatMessage.java
+++ b/src/main/java/com/danielagapov/spawn/Models/ChatMessage/ChatMessage.java
@@ -2,15 +2,55 @@ package com.danielagapov.spawn.Models.ChatMessage;
 
 import jakarta.persistence.Entity;
 import jakarta.persistence.Id;
+import jakarta.persistence.GeneratedValue;
 import java.io.Serializable;
 
 @Entity
-public record ChatMessage (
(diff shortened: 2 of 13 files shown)


src/main/java/com/danielagapov/spawn/Models/ChatMessage/ChatMessage.java
```diff
@@ -2,15 +2,55 @@
...
 @Entity
-public record ChatMessage (
-        @Id
-        Long id,
-        String timestamp, // TODO: investigate data type later
-        Long userSenderId,
-        String content,
-        Long eventId
-) implements Serializable {
+class ChatMessage implements Serializable {
```
Daggerpov, line 9: Your new model classes all look good. Good job

src/main/java/com/danielagapov/spawn/Models/FriendTag/FriendTag.java
```diff
@@ -1,14 +1,45 @@
...
+
+        public Long getId() {
+                return id;
+        }
+
+        public void setId(Long id) {
+                this.id = id;
+        }
+
+        private String color; // TODO: investigate data type later
```
Daggerpov, line 44: What is this variable doing down here? lol, could you move it to be along with the rest of them please?

src/main/java/com/danielagapov/spawn/Models/FriendTag/FriendTag.java
```diff
@@ -1,14 +1,45 @@
...
+
+        public void setDisplayName(String displayName) {
+                this.displayName = displayName;
+        }
+
+        public Long getId() {
+                return id;
+        }
+
+        public void setId(Long id) {
```
Daggerpov, line 40: Do we need a setter for id, if it's generated? What are your thoughts on this? I'm not entirely sure.

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -44,4 +44,10 @@ public String getMockEndpoint() {
     public Event createEvent(@RequestBody Event newEvent) {
         return eventService.saveEvent(newEvent);
     }
+
+    // full path: /api/v1/events/{id}
+    @PutMapping("{id}")
+    public Event replaceEvent(@RequestBody Event newEvent, @PathVariable Long id) {
```
Daggerpov, line 50: Proper naming and parameters conventions; great job ✅

Daggerpov: I wasn't previously familiar with `PUT` endpoints, since I've mostly just worked with GET & POST, but after seeing this PR, I dug online to make sure you're following good practices.

src/main/java/com/danielagapov/spawn/Services/FriendTag/IFriendTagService.java
```diff
@@ -7,5 +7,6 @@ public interface IFriendTagService {
     public List<FriendTag> getAllFriendTags();
     public FriendTag getFriendTagById(Long id);
     public List<FriendTag> getFriendTagsByTagId(Long tagId);
-    public FriendTag saveFriendTag(FriendTag event);
+    public FriendTag saveFriendTag(FriendTag friendTag);
```
Daggerpov, line 10: Good catch here

src/main/java/com/danielagapov/spawn/Services/FriendTag/FriendTagService.java
```diff
@@ -49,4 +49,14 @@ public FriendTag saveFriendTag(FriendTag friendTag) {
             throw new BaseSaveException("Failed to save friendTag: " + e.getMessage());
         }
     }
+
+    public FriendTag replaceFriendTag(FriendTag newFriendTag, Long id) {
+        return repository.findById(id).map(friendTag -> {
```
Daggerpov, line 54: Btw, this is a suggestion, but another way of writing a method like this that either inserts/updates a thing based on whether it already exists, is to name it `upsert`. I won't block this current PR, but in the future, I'd like if we abstracted this logic out to a method called `upsertFriendTag` that has this finding if/else

Daggerpov: Then, we can re-use it more, and de-clutter these individual methods. But, it's just a styling preference, so it's not a huge deal.

Daggerpov (approved):
Looks good! You've done a great job conforming to the RESTful PUT endpoint practices, with passing in a body and id throughout the endpoint & service. 

I've left some minor code-style comments, but won't stop them from merging this PR in.

src/main/java/com/danielagapov/spawn/Models/FriendTag/FriendTag.java
SeabertYuan: Yeah I think we don't need it I'll remove it because we don't really want code messing with our ids

PR #38, opened 2024-11-20
Removed location from user + adjusted its properties

Daggerpov:
- User didn't need a `lastLocation` property, so I removed that relation
- Location needed proper attributes, so I added: name (what displays with the event), lat & long coordinates

image


evannawfal (approved):
Looks good, longitude and latitude on location are a good call

PR #41, opened 2024-11-21
Db models

evannawfal:
Created the models for all of the tables from the ERD. I had to change some files that interact with these models so let me know if you see anything that looks fishy


Daggerpov (changes requested):
Hey Evan, since you’ve removed the mock models, could you bring the models out of their nested folders? E.g. instead of /Models/Event/Event.java, it can just be /Models/Event.Java now


Will review the rest when I get back.

Daggerpov:
> Hey Evan, since you’ve removed the mock models, could you bring the models out of their nested folders? E.g. instead of /Models/Event/Event.java, it can just be /Models/Event.Java now
> 
> Will review the rest when I get back.

And please make sure building still works (imports for these models should be adjusted automatically by IntelliJ)

src/main/java/com/danielagapov/spawn/Mappers/ChatMessageMapper.java
```diff
@@ -13,19 +15,19 @@ public static ChatMessageDTO toDTO(ChatMessage entity) {
...
-                entity.getUserSenderId(),
+                entity.getUserSender().getId(),
                 entity.getContent(),
-                entity.getEventId()
+                entity.getEvent().getId()
         );
     }
 
-    public static ChatMessage toEntity(ChatMessageDTO dto) {
+    public static ChatMessage toEntity(ChatMessageDTO dto, User userSender, Event event) {
```
Daggerpov, line 24: We shouldn't need to pass in extra arguments to these mapper methods.

Daggerpov: This logic of relations to the event **should be done at the service layer instead.** DTO <-> Entity mappings are just mappings between one DTO to & from its own entity, not inclusive of others.

src/main/java/com/danielagapov/spawn/Mappers/ChatMessageMapper.java
```diff
@@ -35,9 +37,19 @@ public static List<ChatMessageDTO> toDTOList(List<ChatMessage> chatMessages) {
                 .collect(Collectors.toList());
     }
 
-    public static List<ChatMessage> toEntityList(List<ChatMessageDTO> chatMessageDTOs) {
+    public static List<ChatMessage> toEntityList(List<ChatMessageDTO> chatMessageDTOs, List<User> users, List<Event> events) {
         return chatMessageDTOs.stream()
-                .map(ChatMessageMapper::toEntity)
+                .map(dto -> {
```
Daggerpov, line 42: Same as with the individual `toEntity` method, these lists of users and events aren't meant to be in this file, but at the service layer. So, for this method, please change it to, and let's move this logic.

public static List<ChatMessage> toEntityList(List<ChatMessageDTO> chatMessageDTOs) {
    return chatMessageDTOs.stream()
            .map(ChatMessageMapper::toEntity)
            .collect(Collectors.toList());
}

Daggerpov (changes requested):
Please take a look at my comments in `ChatMessageMapper.java`, as they apply to other mapper files as well.

src/main/java/com/danielagapov/spawn/Mappers/ChatMessageMapper.java
Daggerpov: One key advantage is to handle exceptions in the service layer (where they're meant to be), and not including that "responsibility" in these mapper classes.

src/main/java/com/danielagapov/spawn/Mappers/EventMapper.java
```diff
@@ -15,26 +16,23 @@ public static EventDTO toDTO(Event entity) {
                 entity.getTitle(),
                 entity.getStartTime(),
                 entity.getEndTime(),
-                entity.getLocation(),
+                entity.getLocation() != null ? LocationMapper.toDTO(entity.getLocation()) : null, // Map Location to LocationDTO
```
Daggerpov, line 19: Makes sense 👍 Let's stick to this syntax for null-checking in the future, as well, to stay consistent.

src/main/java/com/danielagapov/spawn/Models/UserFriendTagMapping.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Entity
+@Table(
+        name = "user_friend_tag_mapping",
+        uniqueConstraints = @UniqueConstraint(columnNames = {"user_1", "user_2", "friend_tag_id"})
+)
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
```
Daggerpov, line 20: Thanks for using these:

```
@NoArgsConstructor
@AllArgsConstructor
@Getter
@Setter
```

src/main/java/com/danielagapov/spawn/Models/EventTag.java
```diff
@@ -0,0 +1,33 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private UUID id;
+
+    @ManyToOne
+    @JoinColumn(name = "event_id", referencedColumnName = "id", nullable = false)
+    private Event event;
+
+    @ManyToOne
+    @JoinColumn(name = "tag_id", referencedColumnName = "id", nullable = false)
```
Daggerpov, line 31: Seems like we're using `tagId` everywhere instead of `friendTagId`, so it's good you're sticking with that convention 👍

src/main/java/com/danielagapov/spawn/Models/FriendRequests.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+import java.io.Serializable;
+import java.util.UUID;
+
+@Entity
+@Table(
+        name = "friend_requests",
+        uniqueConstraints = @UniqueConstraint(columnNames = {"sender_id", "receiver_id"})
```
Daggerpov, line 16: Great example entity where this unique constraint is very useful. Good job. In other entities, I'm not sure if they're as necessary, but they're fine to keep for principal's sake.

src/main/java/com/danielagapov/spawn/Models/Location.java
```diff
@@ -0,0 +1,27 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class Location implements Serializable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private UUID id;
+
+    @Column(length = 200)
```
Daggerpov, line 22: 255 is more standard (based on # bits), from what I've seen, but this is fine. Not a big deal. 

Keep in mind, at the database level, we don't need to worry about cutting off long text and such, since that's more of a responsibility on the mobile side. 

This would especially be the case if we had both a mobile & web version of the app, in that we don't want to limit the length at the database level, but let each client side make its own config.

src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -0,0 +1,29 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class User implements Serializable {
+        @Id
+        @GeneratedValue
+        private UUID id;
+
+        @Column(nullable = false, unique = true) // Ensures the username is unique and not null
```
Daggerpov, line 22: This might be a tricky one. 

Let's say a user signs in through email & pass, google, or apple -> get to "enter your password" screen
At this point, if they close the app, there wouldn't be a way of us remembering that they signed in, so they'd be prompted again, with this constraint. 

But, maybe that's a concern for the mobile side. Let's keep it this way for now, since I don't want to deal with nullable `username` properties anyway; that'd add unnecessary complexity.

src/main/java/com/danielagapov/spawn/Models/ChatMessage.java
```diff
@@ -0,0 +1,35 @@
...
+
+        @ManyToOne
+        @JoinColumn(name = "user_id", nullable = false)
+        private User userSender;
+
+        @ManyToOne
+        @JoinColumn(name = "event_id", nullable = false)
+        private Event event;
+
+        @Column(length = 1000)
```
Daggerpov, line 33: Let's not limit the content in the database entity layer, if we don't have to, as a general rule-of-thumb. Leave this to the front-end side to determine. Please remove this `length` constraint.

For example, at work, we had a "Comments" field in one of our forms that geotechnical engineers use. The field was restricted to 1000 chars, but they wanted that expanded. It would've been easy if this was configured in the front-end, but since it was at the database level, it required creating a new column and moving the data.

Daggerpov (changes requested):
Requested a few changes:
- Move logic performed in `Mapper` classes to the service layer, from these methods (e.g. `toDTO()`)
- Remove length restrictions at database entity level (e.g. 1000 for `ChatMessage` `content`)

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
```diff
@@ -16,39 +20,54 @@
 
 @Service
 public class ChatMessageService implements IChatMessageService {
-    private final IChatMessageRepository repository;
+    private final IChatMessageRepository chatMessageRepository;
```
Daggerpov, line 23: Fine for now. We might want to consider using the mediator design pattern later on to avoid all of these dependencies.

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
```diff
@@ -16,39 +20,54 @@
...
-            ChatMessage chatMessageEntity = ChatMessageMapper.toEntity(chatMessage);
-            repository.save(chatMessageEntity);
+            // Fetch User and Event entities
+            User userSender = userRepository.findById(chatMessageDTO.userSenderId())
+                    .orElseThrow(() -> new BaseNotFoundException(chatMessageDTO.userSenderId()));
+            Event event = eventRepository.findById(chatMessageDTO.eventId())
+                    .orElseThrow(() -> new BaseNotFoundException(chatMessageDTO.eventId()));
+
+            // Convert DTO to Entity
+            ChatMessage chatMessageEntity = ChatMessageMapper.toEntity(chatMessageDTO, userSender, event);
```
Daggerpov, line 65: As stated earlier, remove function parameters, `userSender` and `event` from this `toEntity()` (and other `Mapper`) methods

Daggerpov: Actually, it's better to keep repository-interfacing in the service layer. What you have is good, upon re-thinking.

Daggerpov (approved):
Great job! I've reviewed thoroughly, and am happy to approve.

I initially left comments to remove arguments to `Mapper` class methods, but realized that it's good you're fetching the, say, locations in the `Service` layer, because that's where we want to handle exceptions anyway.

PR #45, opened 2024-12-08
Updated configurations to use .env

### LINKED ISSUE #43: Host Back-End Online
- [ ] We should be able to re-deploy easily through Git commits to the `main` branch
- [ ] The endpoints should be accessible through the web, not just by running locally

We should investigate how to deploy the back-end. One way is to containerize it through Docker, then host that. 
Otherwise, we could have some service, like [Railway ](https://railway.app/dashboard)host it for us through Git deployments in an easy way

### LINKED ISSUE #44: Host Database Online
Our hosted back-end should be able to communicate with a hosted database, as opposed to a local one, as we've setup now.

evannawfal:
Using .env to set environment variables so that they aren't hardcoded into the app, allowing us to test connections on railway using this repository. Railway connections are also implemented, backend is now hosted online

Resolves #43 
Resolves #44

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 # https://github.com/github/gitignore/blob/main/community/Golang/Go.AllowList.gitignore
 #
 # Binaries for programs and plugins
+.env
 *.exe
 *.exe~
 *.dll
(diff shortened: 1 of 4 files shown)


Daggerpov (approved):
Good practice 👍

PR #46, opened 2024-12-11
GET /users/friendtag/{tag}

OnionFastener:
endpoint for GET users by friend tag


src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -33,6 +34,12 @@ public String getMockEndpoint() {
         return "This is the mock endpoint for users. Everything is working with it.";
     }
 
+    // full path: /api/v1/users/friendtag/{tag}
+    @GetMapping("/friendTag/{tag}")
```
Daggerpov, line 38: Make the var name match the path, please

Daggerpov (changes requested):

PR #47, opened 2024-12-12
Chatmessage likes features

### LINKED ISSUE #43: Host Back-End Online

### LINKED ISSUE #44: Host Database Online

evannawfal:
Implemented endpoints, service methods, and tests to:

- Create a chatMessageLike
- Delete a chatMessageLike
- Get a list of UserDTOs of people who liked a given comment

This PR also contains changes from the db-models PR but I accidentally branched from that one when I made these changes (oops)

Resolves #43 
Resolves #44

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
@@ -3,7 +3,9 @@ package com.danielagapov.spawn.Controllers;
 import com.danielagapov.spawn.DTOs.ChatMessageDTO;
 import com.danielagapov.spawn.Services.ChatMessage.IChatMessageService;
 import org.springframework.web.bind.annotation.*;
+import com.danielagapov.spawn.DTOs.UserDTO;
 
+import java.util.List;
 import java.util.UUID;
 
 @RestController()
@@ -37,6 +39,24 @@ public class ChatMessageController {
     @PostMapping
     public ChatMessageDTO createChatMessage(@RequestBody ChatMessageDTO newChatMessage) {
         return chatMessageService.saveChatMessage(newChatMessage);
+    }
+
+    // full path: /api/v1/chatMessages/{chatMessageId}/likes/{userId}
+    @PostMapping("/{chatMessageId}/likes/{userId}")
+    public void createChatMessageLike(@PathVariable UUID chatMessageId, @PathVariable UUID userId) {
+        chatMessageService.createChatMessageLike(chatMessageId, userId);
+    }
+
+    // full path: /api/v1/chatMessages/{chatMessageId}/likes
+    @GetMapping("/{chatMessageId}/likes")
+    public List<UserDTO> getChatMessageLikes(@PathVariable UUID chatMessageId) {
+        return chatMessageService.getChatMessageLikes(chatMessageId);
+    }
+
+    // full path: /api/v1/chatMessages/{chatMessageId}/likes/{userId}
+    @DeleteMapping("/{chatMessageId}/likes/{userId}")
+    public void deleteChatMessageLike(@PathVariable UUID chatMessageId, @PathVariable UUID userId) {
+        chatMessageService.deleteChatMessageLike(chatMessageId, userId);
     }
 }
 
--- a/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
@@ -1,9 +1,15 @@
 package com.danielagapov.spawn.Controllers;
 
 import com.danielagapov.spawn.DTOs.ChatMessageDTO;
+import com.danielagapov.spawn.Exceptions.Base.BaseDeleteException;
+import com.danielagapov.spawn.Exceptions.Base.BasesNotFoundException;
 import com.danielagapov.spawn.Services.ChatMessage.IChatMessageService;
 import org.springframework.web.bind.annotation.*;
 import com.danielagapov.spawn.DTOs.UserDTO;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+import com.danielagapov.spawn.DTOs.ChatMessageLikesDTO;
 
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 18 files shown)


SeabertYuan (approved):

Daggerpov (changes requested):
Please adjust the endpoints to use query parameters instead of multiple path vars

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
```diff
@@ -55,22 +65,67 @@ public List<ChatMessageDTO> getChatMessagesByTagId(UUID tagId) {
...
+        List<ChatMessageLikes> likes = chatMessageLikesRepository.findByChatMessage(chatMessage);
+
+        return likes.stream().map(like -> UserMapper.toDTO(like.getUser())).collect(Collectors.toList());
+    }
+
+    public void deleteChatMessageLike(UUID chatMessageId, UUID userId) {
+        try {
+            boolean exists = chatMessageLikesRepository.existsByChatMessage_IdAndUser_Id(chatMessageId, userId);
+            if (!exists) {
+                throw new BaseDeleteException("Like not found for chatMessageId: " + chatMessageId + " and userId: "
```
Daggerpov, line 121: Split these two types of errors into two proper types of errors, instead of them both being `BaseDeleteException`s with different messages. Maybe: `LikeNotFoundException` and `LikeDeletionException`

src/main/java/com/danielagapov/spawn/Services/ChatMessage/IChatMessageService.java
```diff
@@ -4,10 +4,14 @@
...
+import com.danielagapov.spawn.DTOs.UserDTO;
 
 public interface IChatMessageService {
     public List<ChatMessageDTO> getAllChatMessages();
     public ChatMessageDTO getChatMessageById(UUID id);
     public List<ChatMessageDTO> getChatMessagesByTagId(UUID tagId);
     public ChatMessageDTO saveChatMessage(ChatMessageDTO chatMessage);
+    public void createChatMessageLike(UUID chatMessageId, UUID userId);
+    public List<UserDTO> getChatMessageLikes(UUID chatMessageId);
+    public void deleteChatMessageLike(UUID chatMessageId, UUID userId);
```
Daggerpov, line 16: make this return type `boolean`, as discussed in another comment, to represent whether the deletion was successful

src/main/java/com/danielagapov/spawn/Services/ChatMessage/IChatMessageService.java
```diff
@@ -4,10 +4,14 @@
...
 
 import com.danielagapov.spawn.DTOs.ChatMessageDTO;
+import com.danielagapov.spawn.DTOs.UserDTO;
 
 public interface IChatMessageService {
     public List<ChatMessageDTO> getAllChatMessages();
     public ChatMessageDTO getChatMessageById(UUID id);
     public List<ChatMessageDTO> getChatMessagesByTagId(UUID tagId);
     public ChatMessageDTO saveChatMessage(ChatMessageDTO chatMessage);
+    public void createChatMessageLike(UUID chatMessageId, UUID userId);
```
Daggerpov, line 14: Return `ChatMessageLikeDTO` here instead of void

Daggerpov (changes requested):
Hey Evan, overall, great work! I've left some comments for revisions. Some are simply styling, but a majority are core functionality improvements to how our API would help the mobile side. For example, with return types and responses from the controllers & services, would greatly help in communicating to the front-end what's happening.

src/test/java/com/danielagapov/spawn/ChatMessageTests/ChatMessageLikesTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+    @Mock
+    private IUserRepository userRepository;
+
+    @InjectMocks
+    private ChatMessageService chatMessageService;
+
+    @BeforeEach
+    void setup() {
+        MockitoAnnotations.openMocks(this); // Initialize mocks
```
Daggerpov, line 47: Great job mocking the repo & service classes 👌

src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
```diff
@@ -38,6 +40,24 @@ public String getMockEndpoint() {
...
+
+    // full path: /api/v1/chatMessages/{chatMessageId}/likes
+    @GetMapping("/{chatMessageId}/likes")
+    public List<UserDTO> getChatMessageLikes(@PathVariable UUID chatMessageId) {
+        return chatMessageService.getChatMessageLikes(chatMessageId);
+    }
+
+    // full path: /api/v1/chatMessages/{chatMessageId}/likes/{userId}
+    @DeleteMapping("/{chatMessageId}/likes/{userId}")
+    public void deleteChatMessageLike(@PathVariable UUID chatMessageId, @PathVariable UUID userId) {
```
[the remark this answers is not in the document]
SeabertYuan: Addressed this in my recent commit. Did not change void since I wrapped the entire controller body in try catch.

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
SeabertYuan: I abstracted `BaseNotFoundException` rather than making new exceptions since I noticed a lot of other sections were also using it.

src/main/java/com/danielagapov/spawn/Services/ChatMessage/IChatMessageService.java
SeabertYuan: Addressed in reply to the other comment.

SeabertYuan:
I should add that the `ChatMessageLikesDTO` and the corresponding mapper are stubs for now.

src/main/java/com/danielagapov/spawn/DTOs/ChatMessageLikesDTO.java
```diff
@@ -0,0 +1,7 @@
+package com.danielagapov.spawn.DTOs;
+
+import java.io.Serializable;
+import java.util.UUID;
+
+public record ChatMessageLikesDTO (
```
Daggerpov, line 6: Let's have this extend the entity, `ChatMessageLikes`. Then, if the DTO needs specific data fields, we can add them. This is how we can solve the issue of having this empty DTO, which can lead to client-side confusion of getting nothing back.

Daggerpov (approved):
Let's merge this in, since it's been in review for a while. Then, let's address the remaining comments in a separate PR. I agree with Seabert that we shouldn't leave this in review for too long, since it can cause headache with merge conflicts.

PR #50, opened 2024-12-21
DTO Adjustments + Streamlining User <-> Friend Tag Relationships

Daggerpov:
These changes adjust for the discrepancies between our back-end DTOs and what's needed on the mobile side. 

I put logic for mapping between multi-entities to 1 DTO, for example, with a User and its friends -> in service classes, like `UserService`. I kept the Mapper files lean by simply supplying arguments, like `friends` to a user DTO method, for example.

@SeabertYuan and I made lots of changes to these service files to account for changes to user<-> friend tag relationships: `UserService`, `ChatMessageService`, `FriendTagService` + created `FriendRequestService`


SeabertYuan:
Everything runs now :)

src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -18,12 +20,13 @@ public class User implements Serializable {
         @Id
         @GeneratedValue
         private UUID id;
-
+        private List<User> friends;
```
Daggerpov, line 23: Leave this out of the entity. It should only be in the DTO. Refer to the README's DTO drawing, as this case happens to exactly match it. 

Basically, `friends` as a list can't actually be put directly into SQL, which is what this `@Entity` tag does behind the hood with Spring Data JPA to a MySQL table. 

So, while we need the `friends` property in the DTO, we need to fetch that appropriately in the mapper files, not from the entity directly.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -46,17 +42,11 @@ public UserDTO getUserById(UUID id) {
...
-            for (UserFriendTagMapping mapping : mappings) {
-                users.add(mapping.getUser2());
-            }
-            return UserMapper.toDTOList(users);
+            return UserMapper.toDTOList(repository.findAll());
         } catch (DataAccessException e) {
-            throw new DatabaseException("Failed to get users by tag ID: " + e.getMessage(), e);
-        } catch (Exception e) {
-            throw new ApplicationException("Unexpected error occurred while getting users by tag ID: " + e.getMessage(), e);
+            throw new RuntimeException("Error retrieving users", e);
```
Daggerpov, line 49: Could you please not undo the work Shane did in #57 in using the `DatabaseException` and `ApplicationException`

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov, line 23: Leave this out of the entity. It should only be in the DTO. Refer to the README's DTO drawing, as this case happens to exactly match it. 

Basically, `friends` as a list can't actually be put directly into SQL, which is what this `@Entity` tag does behind the hood with Spring Data JPA to a MySQL table. 

So, while we need the `friends` property in the DTO, we need to fetch that appropriately in the mapper files, not from the entity directly.

Daggerpov (commented):
@SeabertYuan

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov: As a rule-of-thumb, you should never have `List`s in `@Entity` objects, as those can't be translated into SQL

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
SeabertYuan: Oops my bad, careless merging.

src/main/java/com/danielagapov/spawn/Models/User.java
SeabertYuan: Upon further research, I think that the easiest solution is to use a series of `@JoinColumn` and `@ManyToOne` (and its variants) annotations. This will join the tables allowing us to directly have collections in our entities. The problem is that I have no solid grasp of how these join operations work and we have no testing infrastructure in place to ensure that these annotations will behave as we expect granted that we annotate our fields correctly.

SeabertYuan:
This commit is honestly very experimental and I'm not sure if we should merge this into main yet. Like I stated above, I think the best solution is to figure out the annotations.

This commit uses the proposed query method as suggested in issue #63 to build DTO's, I think it's not the most elegant solution if we can get the annotations working.
1. I don't like the coupling between the mappers and the repositories
2. It's more convenient for everything if we directly store collections in the models
3. I'm sure there's performance penalties for the way that I'm currently doing things

src/main/java/com/danielagapov/spawn/Repositories/IFriendTagRepository.java
```diff
@@ -3,10 +3,14 @@
...
+import org.springframework.data.jpa.repository.Query;
 
 import java.util.UUID;
+import java.util.List;
 
 @Repository
 public interface IFriendTagRepository extends JpaRepository<FriendTag, UUID> {
     // The JpaRepository interface already includes methods like save() and findById()
+    @Query(value = "SELECT * FROM id WHERE owner_id = ?1", nativeQuery = true)
+    List<UUID> findTagsByOwner(UUID ownerId);
```
Daggerpov, line 15: Change to: `List<FriendTag> findByOwnerId(UUID ownerId);` for JPA naming conventions.

Daggerpov: Also, notice how I changed the return type to be `FriendTag`s instead of `UUID`s, because that's what that SQL is returning. If you mean to return ids then adjust the SQL to not return the row, but that id property you're looking for

src/main/java/com/danielagapov/spawn/Repositories/IUserFriendTagRepository.java
```diff
@@ -1,10 +1,16 @@
...
+import org.springframework.data.jpa.repository.Query;
 
 import java.util.UUID;
+import java.util.List;
 
 @Repository
-public interface IUserFriendTagRepository extends JpaRepository<UserFriendTag, UUID> { }
+public interface IUserFriendTagRepository extends JpaRepository<UserFriendTag, UUID> {
+    @Query(value = "SELECT * FROM user_id WHERE friend_tag_id = ?1", nativeQuery = true)
+    List<UUID> findFriendsByTagId(UUID tagId);
```
Daggerpov, line 15: Change name to `findFriendIdsByTagId`

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -46,13 +44,9 @@ public UserDTO getUserById(UUID id) {
     }
 
     public List<UserDTO> getUsersByTagId(UUID tagId) {
```
Daggerpov, line 46: Just a note for this file, based on the logic I wanted you to move out of the mapper files, these service classes is where the logic shouuld be put, when you're calling the `.toDTO()` or `.toDTOList()` methods

Daggerpov (commented):
Lots of changes to be make, primarily to:

- [x] Moving logic from mapper files into service classes, for things such as getting a `User`'s `friends`
- [ ] Not touching entities, but aggregating properties like a `User`'s `friends` only for the DTO

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Seems to be fixed now. Resolving

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov: I disagree with this approach. I think we shouldn't have to touch the entities at all, actually, and that'd be easiest. Instead, we can aggregate related entities in the service classes, to then put into the DTO, with arguments. Please see my related comments in your mapping files, like in `FriendTagMapper.java`, for example.

Daggerpov:
> This commit is honestly very experimental and I'm not sure if we should merge this into main yet. Like I stated above, I think the best solution is to figure out the annotations.
> 
> This commit uses the proposed query method as suggested in issue #63 to build DTO's, I think it's not the most elegant solution if we can get the annotations working.
> 
> 1. I don't like the coupling between the mappers and the repositories
> 2. It's more convenient for everything if we directly store collections in the models
> 3. I'm sure there's performance penalties for the way that I'm currently doing things

1. I've addressed how you can fix that in comments on your mapping files (essentially, moving to service classes, and adding arguments to the `.toDTO()` and `.toDTOList()` methods
2. Addressed this in [this comment](https://github.com/Daggerpov/Spawn-App-Back-End/pull/50#discussion_r1899148569)
3. Yes, joining at the database level is more costly than in the service class layer. So, that's another added bonus of switching over to my approach.

I do really want to thank you, though, @SeabertYuan , for investigating this far and experimenting. It's great that you took it upon yourself to figure out a solution.

src/main/java/com/danielagapov/spawn/Models/User.java
SeabertYuan: Sounds good, I was messing around with this all of yesterday and I couldn't get it to work because of the cyclical dependencies when converting to DTO or to entity. It turns out a lot of this commit you've reviewed doesn't even work when I was testing :/

SeabertYuan:
Finally, I think this PR is just about done now. I've made the requested changes (I hope). There is one kink though, `FriendTagDTO` cannot have `UserDTO`'s in them otherwise we end up with cycles when attempting to map a `UserDTO`. I think the easiest solution is just to use UUID's, and let the mobile side request the specific DTO when it actually needs it.

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov: Done; changed it all myself. Please take a look at how I did it, @SeabertYuan

PR #51, opened 2024-12-24
Create/Save a new Friend Request

Shane-M-1:
I did this without the use of a DTO and instead, this returns a friend request directly. I didn't think a DTO would be necessary considering the simplicity of the FriendRequests model, but let me know if that should change!

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -1,6 +1,7 @@
 package com.danielagapov.spawn.Controllers;
 
 import com.danielagapov.spawn.DTOs.UserDTO;
+import com.danielagapov.spawn.Models.FriendRequests;
 import com.danielagapov.spawn.Services.User.IUserService;
 import org.springframework.web.bind.annotation.*;
 
@@ -51,4 +52,11 @@ public class UserController {
     public UserDTO replaceUser(@RequestBody UserDTO newUser, @PathVariable UUID id) {
         return userService.replaceUser(newUser, id);
     }
+
+    // full path: /api/v1/user/friendReq
+    @PostMapping("friendReq")
+    public FriendRequests createFriendRequest(@RequestBody FriendRequests friendReq) {
+        return userService.saveFriendRequest(friendReq);
+    }
+
 }
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -53,8 +53,8 @@ public class UserController {
         return userService.replaceUser(newUser, id);
     }
 
-    // full path: /api/v1/user/friendReq
-    @PostMapping("friendReq")
+    // full path: /api/v1/user/friend-request
+    @PostMapping("friend-request")
     public FriendRequests createFriendRequest(@RequestBody FriendRequests friendReq) {
         return userService.saveFriendRequest(friendReq);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -51,4 +52,11 @@ public UserDTO createUser(@RequestBody UserDTO newUser) {
     public UserDTO replaceUser(@RequestBody UserDTO newUser, @PathVariable UUID id) {
         return userService.replaceUser(newUser, id);
     }
+
+    // full path: /api/v1/user/friendReq
+    @PostMapping("friendReq")
```
SeabertYuan, line 57: Going off of the resource @Daggerpov sent on Slack, `/user/{id}/friend-requests` would follow REST URI conventions closer.

SeabertYuan (commented):
From a design standpoint I think it's better to use DTO's for consistency but I think this is something we should discuss with @Daggerpov .

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
Shane-M-1: Good point, I'll revisit that site.

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -11,4 +12,5 @@ public interface IUserService {
     public List<UserDTO> getUsersByTagId(UUID tagId);
     public UserDTO saveUser(UserDTO user);
     public UserDTO replaceUser(UserDTO user, UUID id);
+    public FriendRequests saveFriendRequest(FriendRequests friendRequest);
```
Daggerpov, line 15: Please make a DTO, as discussed on Slack, that can simply inherit (`extends`) from `FriendRequests` for now.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -70,4 +74,13 @@ public UserDTO replaceUser(UserDTO newUser, UUID id) {
             return UserMapper.toDTO(userEntity);
         });
     }
+
+    public FriendRequests saveFriendRequest(FriendRequests friendRequest) {
+        try {
+            friendRequestsRepository.save(friendRequest);
+            return friendRequest;
```
Daggerpov, line 81: Once you make the entity -> DTO change, please save the entity to the database, and return the DTO to the controller (for the client [mobile]).

Daggerpov (changes requested):
Great work. Just a couple of comments regarding the use of DTOs.

src/main/java/com/danielagapov/spawn/DTOs/FriendRequestDTO.java
```diff
@@ -0,0 +1,9 @@
+package com.danielagapov.spawn.DTOs;
+
+import com.danielagapov.spawn.Models.User;
+
+import java.io.Serializable;
+import java.util.UUID;
+
+public record FriendRequestDTO(UUID id, User sender, User receiver) implements Serializable {
```
Daggerpov, line 8: Hey, looks great! Just one thing, could you switch these `User` objects to `UserDTO`? Apart from that, everything else looks completely solid, so I'll approve right afterwards.

Daggerpov (changes requested):
Looks great! Just `User` -> `UserDTO` in the `FriendRequestDTO`

Daggerpov (approved):

PR #55, opened 2024-12-25
Entity type for errors

Daggerpov:
- Created `EntityType` to use as an enum, rather than passing in the type as a string to errors.


Shane-M-1 (approved):
Looks good

PR #56, opened 2024-12-25
Using composite keys instead of generated ones for join tables

Daggerpov:
Created composite key classes, to then use in place of generated ones in our join table models


Daggerpov:
@evannawfal Not sure if this requires changes in our database. Could you checkout this branch and see if any additional changes need to be made to make this work?

Daggerpov:
Same thing should probably be done for more files. Feel free to push to this branch/PR. For example, in this file:

evannawfal (approved):
Good change, this makes the database structure cleaner and ensures the relationships are handled more naturally

PR #57, opened 2024-12-25
Get Users by Friend Tag ID

Shane-M-1:
Added the functionality for getting users by friend tag to UserService.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/User/UserService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/User/UserService.java
@@ -2,9 +2,7 @@ package com.danielagapov.spawn.Services.User;
 
 import com.danielagapov.spawn.DTOs.FriendRequestDTO;
 import com.danielagapov.spawn.DTOs.UserDTO;
-import com.danielagapov.spawn.Exceptions.Base.BaseNotFoundException;
-import com.danielagapov.spawn.Exceptions.Base.BaseSaveException;
-import com.danielagapov.spawn.Exceptions.Base.BasesNotFoundException;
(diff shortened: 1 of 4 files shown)


src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -39,12 +46,12 @@ public UserDTO getUserById(UUID id) {
     }
 
     public List<UserDTO> getUsersByTagId(UUID tagId) {
-        // TODO: change this logic later, once tags are setup.
-        try {
-            return UserMapper.toDTOList(repository.findAll());
-        } catch (DataAccessException e) {
-            throw new RuntimeException("Error retrieving users", e);
+        List<UserFriendTagMapping> mappings = userTagMappingRepository.findByFriendTagId(tagId);
```
Daggerpov, line 49: this could throw exceptions

Daggerpov: So, I don't think it's a problem if there aren't any found -> we don't need to catch something like `EntityNotFoundException`, but, we should surround it with a try-catch with these, probably:

```
} catch (DataAccessException e) {
        // Handle database-related issues
        throw new DatabaseException("Error accessing data for tag ID: " + tagId, e);
    } catch (Exception e) {
        // Handle unexpected issues
        throw new ApplicationException("An unexpected error occurred while fetching users by tag ID.", e);
    }
```

Daggerpov (approved):
Looks good. Moving forward, we can use these custom exceptions, `ApplicationException` and `DatabaseException` in more places, for consistency

Issue #24, opened 2024-11-19
Endpoints: Controllers & Stub Service Methods


Daggerpov:
## Events

- [x] - POST ✅
- [x] - PUT event by event id
    - Overwrite its info
- [x] - GET events by user id, that they’ve been invited to ✅, by @Shane-M-1 : https://github.com/Daggerpov/Spawn-App-Back-End/pull/23
    - Deal with invited to logic later, but for now:
        - Setup endpoint
        - Setup service class, which uses `IEventRepository.java`    
- [x] - GET events by Friend Tag id
- [x] - DELETE event by event id

## Chat Messages

- [x] - GET chat messages given event id
     - These chat messages will compose of users
- [x] - POST a chat message
- [x] - DELETE chat message by message id

## Friend Tags

- [x] - GET Friend Tags given user id
- [x] - PUT Friend Tag given friend tag id
    - e.g. edit name or colour
- [x] - POST Friend Tag given user id and friend tag id
    - e.g. add a user to an existing friend tag
- [x] - DELETE Friend Tag by friend tag id

## User

- [x] - PUT user by user id
     - Overwrite its info
         - For example, with profile image, username, etc. 
- [x] - GET user’s friends by user id
- [x] - POST a friend request, given two user ids (sender, receiver)
- [x] - GET incoming friend requests
- [x] - GET users participating in given id
- [x] - GET user by id ✅
- [x] - POST user (sign-up) ✅
- [x] - GET users by tag
- [x] - DELETE a user's friend by friend id
- [x] - DELETE User by user id
- [x] - GET recommended friends by user id
     - within service class:
         - → get user’s friends
         - → get user’s friends friends’
         - → compare which have the most mutuals

Daggerpov:
Let's close this issue once #73 merges

PR #67, opened 2025-01-03
Adding endpoints

SeabertYuan:
Checked off some endpoints related to `friendTags` and `Users` for #24.


diagrams.excalidraw
```diff
@@ -1,5759 +1,10 @@
 {
   "type": "excalidraw",
   "version": 2,
```
Daggerpov, line 3: Please undo your changes to this file; I added it as a sort of back-up for the excalidraw file, since my free version account doesn't allow saving.

src/main/java/com/danielagapov/spawn/Controllers/FriendTagController.java
```diff
@@ -1,7 +1,7 @@
 package com.danielagapov.spawn.Controllers;
 
 import com.danielagapov.spawn.DTOs.FriendTagDTO;
-import com.danielagapov.spawn.Exceptions.Base.BaseNotFoundException;
+import com.danielagapov.spawn.Exceptions.Base.*;
```
Daggerpov, line 4: `.*` is bad practice for imports, since it'll usually take more processing than necessary. Also, it makes it harder to track down which imports are being used and can be removed, in addition to dependency graphs. To illustrate this, you'll notice if you use IntelliJ's "Optimize Imports" command (accessible through "Ctrl+Shift+P", it will always reduce to the fewest imports possible, meaning it'll convert .* -> only the imports necessary. Not a big deal in our project, but in industry, it can be.

src/main/java/com/danielagapov/spawn/Controllers/FriendTagController.java
```diff
@@ -64,6 +64,21 @@ public ResponseEntity<Void> deleteFriendTag(@PathVariable UUID id) {
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
         }
     }
+
+    // full path: /api/v1/friendTags/{id}?userId=userId
+    @PostMapping("{id}")
+    public ResponseEntity<Void> addUserToFriendTag(@PathVariable UUID id, @RequestParam UUID userId) {
```
Daggerpov, line 70: Good stuff!

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -35,6 +35,12 @@ public UserDTO getUser(@PathVariable UUID id) {
         return userService.getUserById(id);
     }
 
+    // full path: /api/v1/users/{id}/friends
+    @GetMapping("{id}/friends")
+    public List<UserDTO> getUserFriends(@PathVariable UUID id) {
```
Daggerpov, line 40: Very useful new endpoint, and good conventions with the URL!

src/main/java/com/danielagapov/spawn/Services/FriendTag/FriendTagService.java
```diff
@@ -95,4 +105,24 @@ public boolean deleteFriendTagById(UUID id) {
             return false;
         }
     }
+
+    public void saveUserToFriendTag(UUID id, UUID userId) {
+        if (!repository.existsById(id)) {
+            throw new BaseNotFoundException(id);
```
Daggerpov, line 111: Hey Seabert, I know this isn't part of your PR, but as a small thing: could we add an `EntityType` argument to `BaseNotFoundException`, similar to what's found in `EntityAlreadyExistsException`? It'd be useful in this case, because we're throwing the same exception, with two different entity types, so we'd want to be able to distinguish them.

Daggerpov (changes requested):
I just added a couple of very minor comments. Great PR!

diagrams.excalidraw
SeabertYuan: Shoot I accidentally opened it and didn't realize that changed anything sorry!

Daggerpov (approved):
Great work! Thanks for implementing the requested changes.

Issue #58, opened 2024-12-25
Add Documentation to Model Classes


Daggerpov:
Just as simple as adding comments before the classes, explaining what the entity/table is there for, and maybe what the different `Id` properties are there for.

PR #69, opened 2025-01-03
added documentation to User, UserFriendTag, and FriendTag models

SeabertYuan:
Addressed part of #58.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Models/FriendTag.java
+++ b/src/main/java/com/danielagapov/spawn/Models/FriendTag.java
@@ -14,7 +14,8 @@ import java.util.UUID;
 /*
  * Represents a unique FriendTag entity. Each FriendTag entity is crucially
  * associated to one User by the ownerId field. Furthermore, each User contains
- * at least one FriendTag.
+ * at least one FriendTag, being the "Everyone" friend tag that is added by
+ * default.
  */
 @Entity
 // these two annotations are in place of writing out constructors manually (for readability):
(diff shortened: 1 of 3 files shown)


src/main/java/com/danielagapov/spawn/Models/FriendTag.java
```diff
@@ -11,6 +11,11 @@
 import java.io.Serializable;
 import java.util.UUID;
 
+/*
+ * Represents a unique FriendTag entity. Each FriendTag entity is crucially
+ * associated to one User by the ownerId field. Furthermore, each User contains
+ * at least one FriendTag.
```
Daggerpov, line 17:
```suggestion
 * at least one FriendTag, being the "Everyone" friend tag that is added by default.
```

Daggerpov: Btw, we should actually implement this, so I'll make a new issue for that, so that the "Everyone" friend tag is actually added by default upon new user creation.

src/main/java/com/danielagapov/spawn/Models/UserFriendTag.java
```diff
@@ -9,6 +9,11 @@
 import java.io.Serializable;
 import java.util.UUID;
 
+/*
+ * A table consisting of many-to-one relationships of a User's Id and a
+ * FriendTag's Id. This table is queried to find a user's association to a
+ * FriendTag.
```
Daggerpov, line 15: Maybe give an example with names, like "If Seabert creates a friend tag called 'Spawn Friends,' then his friends' user ids, for example, Daniel's user id would be in the `userId` column of this table, while the `friendTagId` would represent the friend tag that contains his user id as its `ownerId` value in the `FriendTag` table."

Daggerpov (changes requested):
Great stuff! Just requested a bit more elaborating

Daggerpov (approved):
Great job with the explanation, thanks!

Issue #70, opened 2025-01-04
Create "Everyone" Friend Tag Upon User Creation + Update Upon Friend Addition


Daggerpov:
Now that we don't have a `UserFriends` table, per @SeabertYuan's changes with #63 , we should be creating an "Everyone" friend tag when a new user is created through our post endpoint in `UserController`. In addition, we need to make sure that every new friend through our controller file is also added to this "Everyone" friend tag. 

Crucially, this work shouldn't actually be done in the Controller level, but at the service level, as nothing in the controller files should change.

Daggerpov:
@SeabertYuan Please make the "Everyone" tag have the colour "#1D3D3D" before closing this issue

Daggerpov:
> @SeabertYuan Please make the "Everyone" tag have the colour "#1D3D3D" before closing this issue

Done by pushing to your PR

PR #72, opened 2025-01-05
Friend user creation

SeabertYuan:
This addresses #70 . I had to update the `User` model to guarantee that we can always find our "everyone" tag. Hoping to get this merged and I will open a separate issue about linking it to friend requests, resolving the DTO conversion issues, and guaranteeing that the `UserDTO` will always have it's first `FriendTagDTO` be the "everyone" tag.


src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -31,4 +33,5 @@ public class User implements Serializable {
         private String lastName;
         private String bio;
         private String email;
+        private UUID allFriends;
```
Daggerpov, line 36: Please don't include these kinds of fields in entities, just in DTOs. I think I've explained this, but 'extra' information on top of entities that are useful for the API calls between our mobile<-> back-end should only be included in the dto parameters, and populated in our service classes, as I did in #50 .

entity-relationship-diagram.dbml
```diff
@@ -6,6 +6,7 @@ Table User {
     lastName String
     bio String
     email String
+    allFriends UUID
```
Daggerpov, line 9: This shouldn't be in the ERD, due to what I commented in the `User.java` file. Basically, we only want this field in the DTO, not the entity. As such, the Entity Relationship Diagram doesn't need to include it.

src/main/java/com/danielagapov/spawn/Mappers/UserMapper.java
```diff
@@ -33,7 +33,8 @@ public static User toEntity(UserDTO dto) {
                 dto.firstName(),
                 dto.lastName(),
                 dto.bio(),
-                dto.email()
+                dto.email(),
+                dto.friendTags().get(0).id()
```
Daggerpov, line 37: This field shouldn't be in the entity, but DTO, so please adjust this mapping to have it the other way around.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -92,15 +92,17 @@ public UserDTO saveUser(UserDTO user) {
...
-            List<UserDTO> friends = getFriendsByFriendTagId(userEntity.getId());
-            List<FriendTagDTO> friendTags = friendTagService.getFriendTagsByOwnerId(userEntity.getId());
+            // TODO: resolve circular entity-DTO conversions
+            FriendTagDTO everyoneTag = new FriendTagDTO(null, "everyone", "#ffffff", user, List.of());
+            UUID everyoneTagId = friendTagService.saveFriendTag(everyoneTag).id(); // id is generated when saving
 
             userEntity = repository.findById(userEntity.getId()).orElseThrow(() ->
                     new BaseSaveException("Failed to retrieve saved user"));
 
+            userEntity.setAllFriends(everyoneTagId);
```
Daggerpov, line 102: Remove this line, as it should be for the DTO, not entity

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -174,4 +176,9 @@ public List<UserDTO> getFriendsByUserId(UUID userId) {
                 .collect(Collectors.toList());
     }
 
+    public void saveFriendToUser(UUID userId, UUID friendId) {
```
Daggerpov, line 179:
```suggestion
    public void addFriend(UUID userId, UUID friendId) {
```

Daggerpov, line 179: Is `friendId` being used? If not, there might be a bug in this method.

Daggerpov (changes requested):
TL;DR: the `allFriends` property should be renamed to `everyoneFriendTagId` and moved from within the entity to being within the DTO

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov: re-thought this, if you decide to go with approach #3, as discussed on Slack, then I'm fine with keeping this in the entity

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -168,12 +164,21 @@ private List<UserDTO> getUserDTOs() {
 
     public List<UserDTO> getFriendsByUserId(UUID userId) {
         // Get the FriendTags associated with the user (assuming userId represents the owner of friend tags)
-        List<FriendTagDTO> friendTags = friendTagService.getFriendTagsByOwnerId(userId);
+        FriendTag everyoneTag;
+        try {
+            everyoneTag = friendTagRepository.findEveryoneTagByOwnerId(userId);
+        } catch (Exception e) {
+            // if everyoneTag is not found the User doesn't exist
```
Daggerpov: I don't like this logic, since it couples the idea of a user with having an "everyone" tag. Say we decide to change behaviour in the future to do with tags, this makes it harder to pull these two concepts apart. Instead, let's just not throw an error here and let `everyoneTag` be null, and just return `List.of()`

Daggerpov (approved):
Addressed 1 change requested + color & name to match Figma

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Addressed in this commit: https://github.com/Daggerpov/Spawn-App-Back-End/pull/72/commits/010e910b4a9550d7090d1a4ae59235ce1cc9292d

PR #76, opened 2025-01-05
Google Authentication

Shane-M-1:


Daggerpov:
To resolve conflicts:

- [ ] Switch to the `main` branch
- [ ] Pull latest
- [ ] Switch back to your `oauth` branch
- [ ] Merge from `main` into your branch
- [ ] Resolve any conflicts

src/main/resources/application.properties
```diff
@@ -7,3 +7,6 @@ spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
 
 spring.jpa.database-platform=org.hibernate.dialect.MySQLDialect
 spring.jpa.hibernate.ddl-auto=update
+
+spring.security.oauth2.client.registration.google.client-id=822760465266-1dunhm4jgrcg17137rfjo2idu5qefchk.apps.googleusercontent.com
```
Daggerpov, line 11: Is this a sensitive key, or can we keep it here publicly?

src/main/java/com/danielagapov/spawn/Mappers/UserMapper.java
```diff
@@ -24,15 +24,16 @@ public static UserDTO toDTO(User user) {
     public static User toEntity(UserDTO dto) {
         // TODO: enable this once relations are setup within entities:
 
-        User user = new User();
```
Daggerpov, line 27: Great change 👌

Daggerpov: Could you do this in `EventMapper`, `FriendRequestMapper`, and `LocationMapper` as well? They seem to have the same redundant `.setId(dto.id())` syntax

src/main/resources/application.properties
Shane-M-1: Google client ID is publicly available so no need to keep it secret

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
```diff
@@ -0,0 +1,58 @@
...
+     * - family_name
+     * - picture
+     * - email
+     *
+     * @param principal
+     * @return
+     */
+    @RequestMapping("google/sign-in")
+    public ResponseEntity<String> googleSignIn(@AuthenticationPrincipal OAuth2User principal) {
+        String given_name = principal.getAttribute("given_name");
```
Daggerpov, line 39: Could you move this logic into `UserService`, or maybe a new service, `OAuthService`, so we can test it? This is probably the most crucial part of our back-end, so it's the most important to test well, which we can do by mocking the service and writing tests (with ChatGPT).

src/main/resources/application.properties
Daggerpov: Ok, sounds good. GitGuardian also checked that it's fine; I was just a bit tripped up by the name

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
Shane-M-1: Yeah I'll probably drop it into another service class

Daggerpov:
Can either remove the unused endpoints, or mark them with the `@Deprecated` tag, along with a little comment as to why it's deprecated:

https://github.com/spring-projects/spring-boot/wiki/Deprecations

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -0,0 +1,55 @@
...
+            return UserMapper.toDTO(savedEntity, null, null);
+        }
+        return userService.getUserById(mapping.getUser().getId());
+    }
+
+    private UserDTO unpackOAuthUser(OAuth2User oauthUser) {
+        String given_name = oauthUser.getAttribute("given_name");
+        String family_name = oauthUser.getAttribute("family_name");
+        String picture = oauthUser.getAttribute("picture"); // TODO: may need to change once S3 is set
+        String email = oauthUser.getAttribute("email"); // to be used as username
```
Daggerpov, line 50: I get that `username` is a required field, but we'd want the user to specify it in the UI. See these screens I've made for the flow, which this oauth login would be between these 2 screens for: 

https://github.com/Daggerpov/Spawn-App-iOS-SwiftUI/pull/73
https://github.com/Daggerpov/Spawn-App-iOS-SwiftUI/pull/74

Daggerpov: I think we want a separate DTO for this unpacked OAuthUser

Daggerpov (changes requested):
Reviewed all 13 files again; just left one comment about unpacking the OAuthUser and our flow. Since, we want the mobile app user to specify their username, not have it filled in using the email. I think we should make a new DTO for `OAuthUserDTO` for the `unpackOAuthUser()` method, then make a new `UserDTO` later, through a different endpoint, after the username and other info is passed from the client. 

Might be something to discuss further.

Daggerpov (approved):
Approving now, then we can (possibly) address the separate DTO concern later

Issue #59, opened 2024-12-25
Return proper HTTP Responses, depending on Service Class exceptions


Daggerpov:
Please refer to #49 for examples of this, as we should try following this practice for all of our controllers, so that the mobile (client) side isn't left without information as to what happened with the API calls.

Daggerpov:
Done in #79

PR #79, opened 2025-01-08
Controllers: Return Proper HTTP Responses

Daggerpov:
#59 

Simple PR, just doing what's been done to a few endpoints with http response codes to all of them for consistency, and so the mobile side can error-handle more effectively.

- Also, implemented the following methods in `EventService.java` properly: `getEventsByTagId()` & `getEventsByUserId()`

For our controller classes, I’ve just realized we should be returning 200 status codes (HttpStatus.OK) instead of 404 (HttpStatus.NOT_FOUND), based on this answer: https://stackoverflow.com/a/13367198/13368695.

This doesn’t require any changes to how we’re throwing BaseNotFoundException and BasesNotFoundException in our service class layer, so don’t worry about that. It’s just that from those errors, we need to return a different status code, as I’ll adjust for in my current PR, Controllers: Return Proper HTTP Responses #79

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -3,6 +3,7 @@ package com.danielagapov.spawn.Controllers;
 import com.danielagapov.spawn.DTOs.EventDTO;
 import com.danielagapov.spawn.DTOs.UserDTO;
 import com.danielagapov.spawn.Exceptions.Base.BaseNotFoundException;
+import com.danielagapov.spawn.Exceptions.Base.BasesNotFoundException;
 import com.danielagapov.spawn.Services.Event.IEventService;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -26,7 +27,7 @@ public class EventController {
         try {
             return new ResponseEntity<>(eventService.getAllEvents(), HttpStatus.OK);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
@@ -36,9 +37,9 @@ public class EventController {
         try {
             return new ResponseEntity<>(eventService.getEventById(id), HttpStatus.OK);
         } catch (BaseNotFoundException e) {
-            return new ResponseEntity<>(HttpStatus.NOT_FOUND); // Resource not found
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
@@ -50,7 +51,7 @@ public class EventController {
         } catch (BasesNotFoundException e) {
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
@@ -60,7 +61,7 @@ public class EventController {
         try {
             return new ResponseEntity<>(eventService.getEventsByTagId(tagId), HttpStatus.OK);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
@@ -76,7 +77,7 @@ public class EventController {
         try {
             return new ResponseEntity<>(eventService.saveEvent(newEvent), HttpStatus.CREATED);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
@@ -86,9 +87,9 @@ public class EventController {
         try {
             return new ResponseEntity<>(eventService.replaceEvent(newEvent, id), HttpStatus.OK);
         } catch (BaseNotFoundException e) {
-            return new ResponseEntity<>(HttpStatus.NOT_FOUND); // Resource not found
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         } catch (Exception e) {
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR); // Unexpected error
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -22,45 +22,72 @@ public EventController(IEventService eventService) {
...
-    
+
     // full path: /api/v1/events/user/{userId}
     @GetMapping("user/{userId}")
-    public List<EventDTO> getEventsByUserId(@PathVariable UUID userId) {
-        return eventService.getEventsByUserId(userId);
+    public ResponseEntity<List<EventDTO>> getEventsByUserId(@PathVariable UUID userId) {
+        try {
+            return new ResponseEntity<>(eventService.getEventsByUserId(userId), HttpStatus.OK);
+        } catch (Exception e) {
```
Shane-M-1, line 50: do we want to catch BaseNotFoundException here as well so we can throw a 404 if needed, like the other endpoints that search by id?

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -22,45 +22,72 @@ public EventController(IEventService eventService) {
...
     }
 
     // full path: /api/v1/events/friendTag/{tagId}
     @GetMapping("friendTag/{tagId}")
-    public List<EventDTO> getEventsByFriendTagId(@PathVariable UUID tagId) {
-        return eventService.getEventsByTagId(tagId);
+    public ResponseEntity<List<EventDTO>> getEventsByFriendTagId(@PathVariable UUID tagId) {
+        try {
+            return new ResponseEntity<>(eventService.getEventsByTagId(tagId), HttpStatus.OK);
+        } catch (Exception e) {
```
Shane-M-1, line 60: same thing about BaseNotFoundException as before

src/main/java/com/danielagapov/spawn/Controllers/FriendTagController.java
```diff
@@ -22,32 +23,58 @@ public FriendTagController(IFriendTagService friendTagService) {
...
     }
 
     // full path: /api/v1/friendTags
     @PostMapping
-    public FriendTagDTO createFriendTag(@RequestBody FriendTagDTO newFriendTag) {
-        return friendTagService.saveFriendTag(newFriendTag);
+    public ResponseEntity<FriendTagDTO> createFriendTag(@RequestBody FriendTagDTO newFriendTag) {
+        try {
+            return new ResponseEntity<>(friendTagService.saveFriendTag(newFriendTag), HttpStatus.CREATED);
+        } catch (BaseSaveException e) {
```
Shane-M-1, line 61: Is it intentional to catch a BaseSaveException but still perform the same behaviour in the catch block for Exception

Shane-M-1 (changes requested):

src/main/java/com/danielagapov/spawn/Controllers/FriendTagController.java
Shane-M-1: Whoops I cut off the lines that I'm trying to reference but what I'm saying is the BaseSaveException and Exception catch blocks are identical.

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
Daggerpov: Hey Shane, you're right. We just didn't properly implement the `getEventsByUserId()`, so it wasn't throwing anything. I've adjusted that method's implementation, and caught `BasesNotFoundException` here in the controller in this commit: https://github.com/Daggerpov/Spawn-App-Back-End/pull/79/commits/f1e943cfb86b9eda00241126a324b617b24a16e6

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
Daggerpov: Caught now, and implemented the `getEventsByFriendTagId()` method in the service class in these 2 commits: https://github.com/Daggerpov/Spawn-App-Back-End/pull/79/files/1961f7f3c04a698d1e3b2c50483250b0fc3b2678..6938232f3adb35ca5f48ff544d4a406bcef13b01

src/main/java/com/danielagapov/spawn/Controllers/FriendTagController.java
Daggerpov: This is more of a stylistic habit I have for readability to know which exceptions can be thrown and dealt with. You're right that it's currently redundant to catch the `BaseSaveException`, since the `Exception` will do the same thing. But, in the future, let's say we want to deal with `BaseSaveException`s in a certain way, this'll make it easier to already know where they're being dealt with. 

I've changed it to remove the redundant catch block, but added a comment in this commit: https://github.com/Daggerpov/Spawn-App-Back-End/pull/79/commits/c879ef342313db7a56acdc435ce3269fb6c6e95d

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -77,8 +105,16 @@ public EventDTO saveEvent(EventDTO event) {
     }
 
     public List<EventDTO> getEventsByUserId(UUID userId) {
-        // TODO: Add proper filtering logic for events by user
-        return EventMapper.toDTOList(repository.findAll());
+        List<Event> events = repository.findAll()
+                .stream()
+                .filter(event -> event.getCreator().getId().equals(userId))
+                .collect(Collectors.toList());
```
SeabertYuan, line 111: From a performance standpoint is it better to let the database handle this? We could easily have a custom JPQL query such as `SELECT e FROM Event e WHERE e.creator = :userId` in the EventRepository.

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -41,10 +46,33 @@ public EventDTO getEventById(UUID id) {
...
+            // Step 2: Collect all friend user IDs
+            List<UUID> friendIds = friends.stream()
+                    .map(UserDTO::id)
+                    .toList();
+
+            // Step 3: Filter events based on whether their owner is in the list of friend IDs
+            List<Event> allEvents = repository.findAll();
+            List<Event> filteredEvents = allEvents.stream()
+                    .filter(event -> friendIds.contains(event.getCreator().getId()))
+                    .collect(Collectors.toList());
```
SeabertYuan, line 68: Is it more efficient to let the repository handle this with custom queries?

SeabertYuan (approved):
I don't think it's worth blocking these changes, but maybe we can look into using the db queries vs. using java. This can be a separate issue. Great job @Daggerpov !

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: I agree that it'd be a better place to put this logic 👍

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Yup, I agree that this would be better

Daggerpov: Implemented that in this commit: https://github.com/Daggerpov/Spawn-App-Back-End/pull/79/commits/de5061bfa3754d19ca71fa85cc1a45d5d8068cf5

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Implemented that in this commit: https://github.com/Daggerpov/Spawn-App-Back-End/pull/79/commits/de5061bfa3754d19ca71fa85cc1a45d5d8068cf5

Issue #53, opened 2024-12-25
Chat Message Likes Mapping


Daggerpov:
Let's fix this mapping (actually implement it). If it makes things easier, do this after #52 . 

<img width="747" alt="image" src="https://github.com/user-attachments/assets/5455c503-7380-481b-b4b2-5b0ced21436f" />

PR #80, opened 2025-01-08
Filling in ChatMessageLikesDTO and Mappings

### LINKED ISSUE #52: Chat Message Likes DTO + Renaming
Please rename `ChatMessageLikes` to `ChatMessageLike` and either convert the DTO to a class, such that it can inherit as I wrote in the comment, or simply copy the properties from its entity. Thanks!

image

Daggerpov:
Fixed #52 and #53


Shane-M-1 (approved):
Nice

PR #86, opened 2025-01-13
FriendTag Constructor Argument Bug - fixed

evannawfal:
Not sure if the tests were supposed to pass or not, but this doesn't crash during the building stage anymore. Let me know if you want me to rework the fix.


Daggerpov (approved):
Must've been a little bug in the dto from #72 , which once merged with my tests, broke them. Thanks for fixing this, @evannawfal

PR #88, opened 2025-01-14
Changed SecurityConfig to stop 403

Shane-M-1:


Daggerpov (approved):
Tested that 403 isn't happening from post requests now. Thanks!

Issue #83, opened 2025-01-11
Remove friend from “Everyone” tag upon friend removal


Daggerpov:
Forgot to do as part of #72, but should be just as simple as what @SeabertYuan did for when a friend relation is made.

Issue #82, opened 2025-01-10
Additional Friend Endpoints and Service Methods


SeabertYuan:
I'm going to bunch up everything into one issue but in steps. Feel free to edit the method names to better fit convention.

## Establish Endpoints

- [x] rejectFriendRequest
- [x] acceptFriendRequest

## Write Service Methods

- [x] rejectFriendRequest
  - This method should delete the friend request.
- [x] acceptFriendRequest
  - This method should make use of existing functions to add friends bidirectionally then delete the friend request.

## Refactor

This step may not be necessary.
- [ ] Could be better to abstract some of the friend-user functionality in the controller classes

Issue #74, opened 2025-01-05
DTO-Entity cycle conversion


SeabertYuan:
Currently the relationship between Users, and FriendTags leads to cycles when attempting to convert the respective entities into DTOs.

Daggerpov:
Hey @SeabertYuan , is this purely a performance issue, or a bug that's affecting functionality?

SeabertYuan:
It affects functionality. The program will stack overflow whenever entity->DTO conversion is required (I think it crashes bidirectionally but I forgot).

Daggerpov:
Ok, then yea, let's see how you can fix this in a PR. Thanks for catching this!

Daggerpov:
Fixed in #99

PR #90, opened 2025-01-14
Friend user creation

SeabertYuan:
Finished up #83 and #82 as they were closely related, with these completed I can work on #74.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -149,7 +149,6 @@ public class UserController {
     public ResponseEntity<Void> deleteFriendFromUser(@PathVariable UUID id, @RequestParam UUID friendId) {
         try {
             userService.removeFriend(id, friendId);
-            userService.removeFriend(friendId, id);
             return new ResponseEntity<>(HttpStatus.OK);
         } catch (BaseNotFoundException e) {
             System.out.println(e.getMessage());
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -195,6 +195,7 @@ public class UserController {
     public ResponseEntity<Void> deleteFriendFromUser(@PathVariable UUID id, @RequestParam UUID friendId) {
         try {
             userService.removeFriend(id, friendId);
+            userService.removeFriend(friendId, id);
             return new ResponseEntity<>(HttpStatus.OK);
         } catch (BaseNotFoundException e) {
             System.out.println(e.getMessage());
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -144,6 +144,38 @@ public class UserController {
         }
     }
 
+    // full path: /api/v1/users/{id}/friend-requests/accept?friendRequestId=friendRequestId
+    @GetMapping("{id}/friend-requests/accept")
+    // TODO: edit ResponseEntity, id is unused but it makes the path nicer??
+    public ResponseEntity<Void> acceptFriendRequest(@PathVariable UUID id, @RequestParam UUID friendRequestId) {
+        try {
+            friendRequestService.acceptFriendRequest(friendRequestId);
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (BaseNotFoundException e) {
+            System.out.println(e.getMessage());
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (Exception e) {
+            System.out.println(e.getMessage());
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    // full path: /api/v1/users/{id}/friend-requests/reject?friendRequestId=friendRequestId
+    @GetMapping("{id}/friend-requests/accept")
+    // TODO: edit ResponseEntity, id is unused but it makes the path nicer??
+    public ResponseEntity<Void> rejectFriendRequest(@PathVariable UUID id, @RequestParam UUID friendRequestId) {
+        try {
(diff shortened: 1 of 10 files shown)


src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -144,6 +144,38 @@ public ResponseEntity<FriendRequestDTO> createFriendRequest(@RequestBody FriendR
         }
     }
 
+    // full path: /api/v1/users/{id}/friend-requests/accept?friendRequestId=friendRequestId
+    @GetMapping("{id}/friend-requests/accept")
+    // TODO: edit ResponseEntity, id is unused but it makes the path nicer??
+    public ResponseEntity<Void> acceptFriendRequest(@PathVariable UUID id, @RequestParam UUID friendRequestId) {
```
Daggerpov, line 150: 1. Please create a `FriendRequestController` file and move this method there
2. `GetMapping` seems really out-of-place for this action, since we're not actually "getting" anything, so let's change it to a `PutMapping`: `@PutMapping("/{userId}/friend-requests/{friendRequestId}/accept")`
3. Change the request param to just another path param. I've previously said to try avoiding this in place of request params, but in this case, it makes sense, since it's sort of a nested relationship that the user and its friend requests have:`acceptFriendRequest(@PathVariable UUID userId, @PathVariable UUID friendRequestId) {`

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -144,6 +144,38 @@ public ResponseEntity<FriendRequestDTO> createFriendRequest(@RequestBody FriendR
...
+            System.out.println(e.getMessage());
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (Exception e) {
+            System.out.println(e.getMessage());
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    // full path: /api/v1/users/{id}/friend-requests/reject?friendRequestId=friendRequestId
+    @GetMapping("{id}/friend-requests/accept")
```
Daggerpov, line 164: Same rework as the above endpoint please. Also, thanks for including the error logging I recently introduced.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -179,8 +179,13 @@ public void saveFriendToUser(UUID userId, UUID friendId) {
     }
 
     public void removeFriend(UUID userId, UUID friendId) {
```
Daggerpov, line 181: Instead of having to call this method twice to remove bidirectional friends, let's just make this method check to remove both directions of the friend relationship.

Daggerpov: So we don't have to do this, and we can instead call it with just one direction, and the method checks the other direction.

```
userService.removeFriend(id, friendId);
userService.removeFriend(friendId, id);
```

src/main/java/com/danielagapov/spawn/Services/FriendTag/IFriendTagService.java
```diff
@@ -13,4 +13,6 @@ public interface IFriendTagService {
     public FriendTagDTO replaceFriendTag(FriendTagDTO friendTag, UUID tagId);
     public boolean deleteFriendTagById(UUID id);
     public void saveUserToFriendTag(UUID id, UUID userId);
+
+    void deleteUserFromFriendTag(UUID everyoneTagId, UUID friendId);
```
Daggerpov, line 17:
```suggestion
    public void deleteUserFromFriendTag(UUID id, UUID userId);
```
Correct me if I'm wrong, but it seemed strange how this method signature didn't match what's in `FriendTagService`, so I adjusted it

src/main/java/com/danielagapov/spawn/Services/FriendTag/FriendTagService.java
```diff
@@ -125,4 +126,20 @@ public void saveUserToFriendTag(UUID id, UUID userId) {
             throw new BaseSaveException("Failed to save new UserFriendTag");
         }
     }
+
+    // Replaced with direct database requests in uftRepository
```
Daggerpov, line 130: No point of marking service methods as deprecated. Let's simply remove it. `@Deprecated` is mostly used for Controller methods (endpoints), to let external clients know about internal changes. But, since this is a purely internal method, we should definitely remove it and replace any occurrences where it's being called.

Daggerpov: This keeps our code clean, and possibly rids it of dependencies onto deprecated methods. If an old api endpoint was calling this method, then you'd mark the endpoint as deprecated and remove the method, not mark the service method as deprecated. In this case, it's simple, we can just delete it and rework the instances where (if) it's being called.

Daggerpov (changes requested):
Reviewed it all and requested some minor changes. Great work sticking to our conventions and getting this feature done

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
Daggerpov: Thanks for addressing this. Resolving

src/main/java/com/danielagapov/spawn/Services/FriendTag/FriendTagService.java
Daggerpov: Thanks for addressing this. Resolving

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Thanks for addressing this, and also for the `saveFriendToUser()` method. Resolving

Daggerpov (approved):
I'm glad you also applied the changes I requested to related things, like for the bidirectional friend removal, applying that to adding. You also did this by moving an extra endpoint from `UserController` to the new `FriendRequestController`. Thanks!

Issue #77, opened 2025-01-05
New Endpoints: Event Participation Related


Daggerpov:
`EventController.java`:

- [ ] `getEventsInvitedTo()` -> to get events that a user is either participating in, or has been invited to

- [x] `isUserParticipating()` -> to see if a given user is participating in an event
    - [x] path variable of the event id
    - [x] request param of the user id

- [x] `isUserInvited()` -> to see if a given user is invited to an event
    - [x] path variable of the event id
    - [x] request param of the user id

- [x] `toggleParticipation()` -> to toggle a user's participation between invited and participating
    - [x] path variable of the event id
    - [x] request param of the user id

`EventService.java`

- [x] `getUserParticipationStatus()` -> returns enum
    - [x] Create enum called `UserParticipationStatus`, with three cases: `participating` , `invited` and `notInvited`
    - [x] Make sure this enum return type determines whether the first two endpoints, `isUserParticipating()` and `isUserInvited()` return true or false
- [ ] Come up with other methods to actually affect event participation, using the `EventInvited` and `EventParticipants` entities

Daggerpov:
Done in #92

PR #91, opened 2025-01-15
created a new EventUser repository and such

OnionFastener:
replacing EventInvited and EventParticipating repos for (hopefully) easier queries for event participation status aiming to solve some issues with #77

i think for future, the enum will need a third case of "not participating." or we can throw an error after querying


src/main/java/com/danielagapov/spawn/Models/EventUser.java
```diff
@@ -0,0 +1,41 @@
...
+@Table(name = "event_users")
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class EventUser implements Serializable {
+    // @EmbeddedId
+    // private EventParticipantsId id;
+
+    // TODO: switch to embeddedID
```
Daggerpov, line 24: Hey Owen, is this still TODO?

src/main/java/com/danielagapov/spawn/Models/EventUser.java
```diff
@@ -0,0 +1,41 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private UUID id;
+
+    @ManyToOne
```
Daggerpov, line 29: Good job with the relationships. They make sense, since one event can have many users, and one user can participate in many events.

src/main/java/com/danielagapov/spawn/Models/EventUser.java
Daggerpov: I agree that a composite key (`EmbeddedId`) would be better, so could you change it to that? There are examples of me doing this in #56

src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
```diff
@@ -0,0 +1,10 @@
+package com.danielagapov.spawn.Repositories;
+
+import com.danielagapov.spawn.Models.EventUser;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.UUID;
+
+public interface IEventUserRepository extends JpaRepository<EventUser, UUID> {
```
Daggerpov, line 8: This is fine to leave as a stub for now, especially since the `JpaReposoitory` already comes with some default queries. We can add them later, as you mentioned.

src/main/java/com/danielagapov/spawn/Models/EventUser.java
Daggerpov: You'd put it into the `src/main/java/com/danielagapov/spawn/Models/CompositeKeys` folder, like this:

```
@Embeddable
public class EventParticipantsId implements Serializable {
    private UUID eventId;
    private UUID userId;
}
```

Daggerpov:
Hey Owen, since you're replacing the need for the `EventInvited` and `EventParticipants` models, could you please remove those files and their repository files (`IEventUserRepository` & `IEventParticipantsRepository`).

Daggerpov (changes requested):
Hey Owen, these are some great changes. You can go a little further by resolving the comments I wrote, which are mostly extensions of your own TODOs/ideas.

src/main/java/com/danielagapov/spawn/Models/EventUser.java
OnionFastener: im not 100% sure if what i've implemented is correct if you could take a look for me :)

Daggerpov: Looks good. Resolving this comment.

Daggerpov (approved):
I've updated the Entity Relationship Diagram dbml & png files to reflect your changes. 

Great work! It's nice you consolidated the two tables into one `EventUser` one.

PR #92, opened 2025-01-15
Event participating endpoints

OnionFastener:
Resolves endpoints in #77. Did not implement EventUser DTOs


src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -123,4 +124,57 @@ public ResponseEntity<List<UserDTO>> getUsersParticipatingInEvent(@PathVariable
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
+
+    // full path: /api/v1/events/{eventId}/participating?userId={userid}
+    @GetMapping("events/{eventId}?userId/participating?userId={userid}")
```
Daggerpov, line 129:
```suggestion
    @GetMapping("{eventId}?userId/participating?userId={userid}")
```
`events/` is already included in the path, from the top `@RequestMapping("api/v1/events")`

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -155,4 +156,25 @@ public boolean deleteEventById(UUID id) {
...
+    public UserParticipationStatus getUserParticipationStatus(UUID eventId, UUID userId) {
+
+        return UserParticipationStatus.invited;
+
+        // get list of all users at an event -> if no event, throw exception
+
+        // search EventUser for the participation status of given user -> no user, throw notInvited enum
+    }
+
+    public EventDTO inviteUser(UUID eventId, UUID userId) {
```
Daggerpov, line 169:
```suggestion
    public void inviteUser(UUID eventId, UUID userId) {
```
This doesn't need to return anything, just throw a possible exception. Please implement this method as part of your PR.

Daggerpov: (Same with the `participateUser()` method, and their interface function signatures)

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -155,4 +156,25 @@ public boolean deleteEventById(UUID id) {
...
         return List.of();
     }
+
+    public UserParticipationStatus getUserParticipationStatus(UUID eventId, UUID userId) {
+
+        return UserParticipationStatus.invited;
+
+        // get list of all users at an event -> if no event, throw exception
+
+        // search EventUser for the participation status of given user -> no user, throw notInvited enum
```
Daggerpov, line 166: Let's implement this in the current PR. The logic makes sense 👍

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -123,4 +124,57 @@ public ResponseEntity<List<UserDTO>> getUsersParticipatingInEvent(@PathVariable
...
+        } catch (BaseNotFoundException e) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (Exception e) {
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    // full path: /api/v1/events/{eventId}/toggleStatus?userId={userId}
+    @GetMapping("events/{eventId}/toggleStatus?userId={userId}")
+    public ResponseEntity<Void> toggleParticipation(@PathVariable UUID eventId, @RequestParam UUID userId) {
```
Daggerpov, line 162: We want our logic to be in service classes, and the controllers to be very 'lean', only containing try catches for http responses and deal with calling a single service method. 

So, please move this logic into a new service class method called `toggleParticipation()`. You can see examples of this across all of our controller files

Daggerpov: Basically, you can move most of what's in the `try{}` block into that function

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -123,4 +124,57 @@ public ResponseEntity<List<UserDTO>> getUsersParticipatingInEvent(@PathVariable
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
+
+    // full path: /api/v1/events/{eventId}/participating?userId={userid}
+    @GetMapping("events/{eventId}?userId/participating?userId={userid}")
+    public ResponseEntity<Boolean> isUserParticipating(@PathVariable UUID eventId, @RequestParam UUID userId) {
+        try {
+            if (eventService.getUserParticipationStatus(eventId, userId) == UserParticipationStatus.participating) {
+            return new ResponseEntity<>(true, HttpStatus.OK);
```
Daggerpov, line 133:
```suggestion
                return new ResponseEntity<>(true, HttpStatus.OK);
```
Not a big deal here, but could you please make sure your IDE formats on save for the future?

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
Daggerpov: One reason is, that the `participateUser()` function call could result in another `try{} catch{}` block, so we don't want to clutter this file with nested try-catches. Another reason, is that if the underlying logic changes in the future, we won't have to change across all of our accessing controllers (which might one day become out of date as APIs--hence why we name our endpoints `v1/`), but rather, we can just change that implementation in only one service class method

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -123,4 +124,57 @@ public ResponseEntity<List<UserDTO>> getUsersParticipatingInEvent(@PathVariable
...
+    public ResponseEntity<Void> toggleParticipation(@PathVariable UUID eventId, @RequestParam UUID userId) {
+        try {
+            if (eventService.getUserParticipationStatus(eventId, userId) == UserParticipationStatus.invited) {
+                eventService.participateUser(eventId, userId);
+                return new ResponseEntity<>(HttpStatus.OK);
+            } else if (eventService.getUserParticipationStatus(eventId, userId) == UserParticipationStatus.participating) {
+                eventService.inviteUser(eventId, userId);
+                return new ResponseEntity<>(HttpStatus.OK);
+            } else {
+                // TODO: throw something when not invited?
```
Daggerpov, line 171: Yes, bad request/400 error, I believe. Please return an appropriate response entity here

Daggerpov (changes requested):
Hey, I've left some early feedback on this, but it seems like it's still in-progress, so I've marked it as a draft. @OnionFastener, please set it to "ready for review" once you're done.

image

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
Daggerpov: I'd suggest you read https://restfulapi.net/resource-naming/ for a guide on how to name this URL, and I'll let you come up with a different approach. Hint: there should never be a need for two '?' + in these Java `@GetMapping()`s, we can leave those request params out and instead have them in the function as arguments, like `(@RequestParam userId)`, as you can see with some of our other endpoints.

Daggerpov: Also, following your approach of having the `UserParticipationStatus`, you could refactor these two endpoints `isParticipating` and `isInvited` into one that just returns the status enum.

Daggerpov:
I've also added another endpoint, `getEventsInvitedTo()` (to get events that a user is either participating in, or has been invited to), which your service methods should make pretty simple to implement.

Daggerpov:
@OnionFastener Only just thought about this, but we might want an `EventParticipationDTO`, containing: `eventId`, `userId`, and `EventParticipationStatus` (btw, I'd like if you renamed `UserParticipationStatus` -> `EventParticipationStatus`).

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
OnionFastener: i removed the participateUser method entirely as it has no real functionality. toggleParticipation, once implemented should do that job anyway. I think this allows us to use inviteUser more flexibly

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -231,4 +242,98 @@ public boolean deleteEventById(UUID id) {
...
+        return false;
+    }
+
+    public List<EventDTO> EventsInvitedTo(UUID id) {
+        List<EventUser> eventUsers = eventUserRepository.findByUser_Id(id);
+
+        List<Event> events = new ArrayList<>();
+
+        for (EventUser eventUser : eventUsers) {
+            if (eventUser.getUser().getId().equals(id)) {
```
Daggerpov, line 332: I think we may want to filter out the `notInvited` `EventUser` objects here, ensuring they're either invited to or participating in that event

Daggerpov (approved):
Great changes, just two clarifying questions, so I'll approve and we can discuss it before merging

Daggerpov (changes requested):
Let's discuss the two questions I asked before merging

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -123,4 +125,63 @@ public ResponseEntity<List<UserDTO>> getUsersParticipatingInEvent(@PathVariable
...
+        } catch (Exception e) {
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    // full path: /api/v1/events/{userId}/invitedEvents
+    @GetMapping("{userId}/invitedEvents")
+    public ResponseEntity<List<EventDTO>>getEventsInvitedTo(@PathVariable UUID id) {
+        try {
+            return new ResponseEntity<>(eventService.getEventsByUserId(id), HttpStatus.OK);
```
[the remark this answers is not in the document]
OnionFastener: you are absolutely right. I wrote an `getEventsInvitedTo()` function but forgot to use it here. Thanks for pointing it out!

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
OnionFastener: yup! makes sense to me, filtering here just means that they never enter that if and then return false at the end right? works for me i think

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -330,11 +330,16 @@ public boolean toggleParticipation(UUID eventId, UUID userId) {
         return false;
     }
 
-    public List<EventDTO> EventsInvitedTo(UUID id) {
+    public List<EventDTO> getEventsInvitedTo(UUID id) {
         List<EventUser> eventUsers = eventUserRepository.findByUser_Id(id);
 
         List<Event> events = new ArrayList<>();
 
+        if (events.isEmpty()) {
```
Daggerpov, line 338:
```suggestion
        if (eventUsers.isEmpty()) {
```
Typo?

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Yea, exactly

Daggerpov (changes requested):
Typo?

Daggerpov (approved):
Great work!

PR #93, opened 2025-01-16
introduced TempUserDTO

Shane-M-1:
Introduced a TempUserDTO instead of sending back a half-baked User. With this approach, I realized that persisting a TempUser is pointless since it would be almost immediately deleted so there is no entity for a TempUserDTO.

In this updated flow, to differentiate between an existing user logging in and a prospective user creating an account, a status code 201 is sent back when a new user is trying to create an account, and a status code 200 is sent back when an existing user is trying sign in.


src/main/java/com/danielagapov/spawn/DTOs/AbstractUserDTO.java
```diff
@@ -0,0 +1,4 @@
+package com.danielagapov.spawn.DTOs;
+
+public interface AbstractUserDTO {
```
Daggerpov, line 3: Fire

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -17,39 +19,47 @@ public class OAuthService implements IOAuthService {
...
-        UserIdExternalIdMap mapping = externalIdMapRepository.findById(externId).orElse(null);
-        if (mapping == null) {
-            User userEntity = UserMapper.toEntity(user);
-            User savedEntity = userRepository.save(userEntity);
+        return mapping == null ? tempUser : getUserDTO(mapping);
+    }
 
-            UserIdExternalIdMap newMapping = new UserIdExternalIdMap(externId, savedEntity);
-            externalIdMapRepository.save(newMapping);
+    // REQUIRES: userDTO.id() == externId from unpackOAuthUser()
```
Daggerpov, line 36: Could you throw an exception instead of having this "requires"? Better to bullet-proof, than rely on us using the method properly, in the CPSC 210 style, in my opinion

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -17,39 +19,47 @@ public class OAuthService implements IOAuthService {
...
+    public UserDTO makeUser(UserDTO userDTO, String externId) {
+        // user dto -> entity
+        User userEntity = UserMapper.toEntity(userDTO);
+        // save entity
+        userEntity = userRepository.save(userEntity);
```
Daggerpov, line 41: Let's call the `UserService::saveUser()` method instead of saving it ourselves (i.e. doing `.toEntity()` and `.save()` here), because it also has logic related to the everyone friend tag, and we want to keep the logic consistent. In this case, we'd be sending back the dto without the everyone tag, which would be different than the service's `saveUser()` method

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
```diff
@@ -19,10 +21,20 @@ public OAuthController(IOAuthService oauthService) {
 
     // full path: /api/v1/oauth/google/sign-in
     @RequestMapping("google/sign-in")
-    public ResponseEntity<UserDTO> googleSignIn(@AuthenticationPrincipal OAuth2User principal) {
+    public ResponseEntity<AbstractUserDTO> googleSignIn(@AuthenticationPrincipal OAuth2User principal) {
+        try {
+            AbstractUserDTO verifiedUserDTO = oauthService.verifyUser(principal);
+            return verifiedUserDTO instanceof UserDTO ? ResponseEntity.ok().body(verifiedUserDTO) : ResponseEntity.created(URI.create("api/v1/users")).body(verifiedUserDTO);
```
Daggerpov, line 27: Status code (201 - create) makes a lot of sense here, indicating we need to do further action to create the user. This is really smart.

Daggerpov (changes requested):
Only requested a couple of very minor changes. Really great stuff, and clever `AbstractUserDTO` & status code usage!

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
Shane-M-1: In that method, we don't have access to the external id because we aren't persisting the TempUser so I can't make that check for whether to throw an exception. To make this work, mobile should be sending back the same id as is being received in the response to calling /google/sign-in endpoint. It should be almost guaranteed for the requires clause to be true, as long as Mobile doesn't generate their own user ids before sending it to the OAuthController.

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
Shane-M-1: Addressed in most recent push

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
Daggerpov: Ok, thanks for clarifying. We can leave it be

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
Daggerpov: Looks good; thanks

Daggerpov (approved):
Great improvement in the design and flow!

PR #94, opened 2025-01-17
Sophisticated Error Logging

Daggerpov:
- Removed old error logging from controller files, since that should really be done in the service layer
- Created `ILogger` with implementation class, `Logger` that prints out readable error messages, with: file name, method name, and line numbers along with the message
    - The reason for making this an interface and using dependency injection is so we can easily swap out the implementation in the future
    - Chose to make it tagged with the `@Service` stereotype, due to the documentation [here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/stereotype/package-summary.html) aligning well with its use
- Used the `Logger`'s `.log(String message)` method in service class exception handling
- Caught uncaught `Exception`s in service classes to log & throw them


Daggerpov:
Testing on Railway atm

Issue #100, opened 2025-01-21
Additional "Full" DTOs with Sub-Objects


Daggerpov:
Seabert, Owen, and I chatted -> We agree to keep all current DTOs, entities, and mappers the same, to avoid circular dependencies between them.

However, we can go with option #1 of creating new DTOs e.g. FullEventDTO which has the nested objects (i.e. ChatMessage[] , User for creator, and User[] for invited), as we had before. Seabert will make a PR to construct these DTOs within our service classes, composing our ‘core’ entities (e.g. Event). 

- [ ] `FullEventDTO`, containing its properties as objects, not as ids
    - [ ] `locationId` -> `location` (`Location`)
    - [ ] `creatorUserId` -> `creator` (`User`)
    - [ ] ...

Essentially, look back to what we had in our DTOs folder as of this old commit: https://github.com/Daggerpov/Spawn-App-Back-End/tree/baafe5944221145466075992a93896c7c2807dda/src/main/java/com/danielagapov/spawn/DTOs

Daggerpov:
Done in #101 and #102

PR #101, opened 2025-01-21
Creating full dtos

SeabertYuan:
Guessed which fields we wanted to make full. Please feel free to suggest changes and let me know which endpoints I should add. When I have more time I will investigate redundant methods I think that there might be some methods that we'll never end up using that I created with this PR.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -1,6 +1,7 @@
 package com.danielagapov.spawn.Controllers;
 
 import com.danielagapov.spawn.DTOs.EventDTO;
+import com.danielagapov.spawn.DTOs.FullEventDTO;
 import com.danielagapov.spawn.DTOs.UserDTO;
 import com.danielagapov.spawn.Exceptions.Base.BaseNotFoundException;
 import com.danielagapov.spawn.Exceptions.Base.BasesNotFoundException;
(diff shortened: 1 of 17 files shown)


src/main/java/com/danielagapov/spawn/DTOs/FullChatMessageDTO.java
```diff
@@ -0,0 +1,15 @@
...
+import java.io.Serializable;
+import java.time.Instant;
+import java.util.List;
+import java.util.UUID;
+
+public record FullChatMessageDTO(
+        UUID id,
+        String content,
+        Instant timestamp,
+        UserDTO senderUserId,
```
Daggerpov, line 12: Please change the names in your full DTOs to not include `Id`. Mentioned this in the issue #100

src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -43,6 +44,18 @@ public ResponseEntity<EventDTO> getEventById(@PathVariable UUID id) {
         }
     }
 
+    // full path: /api/v1/events/{id}?full=full
+    @GetMapping("{id}")
+    public ResponseEntity<FullEventDTO> getEventById(@PathVariable UUID id, @RequestParam boolean full) {
```
Daggerpov, line 49: I think these endpoint method names `getEventById` and `getFullEventById` are mixed up, considering the object you're returning

src/main/java/com/danielagapov/spawn/Services/Location/LocationService.java
```diff
@@ -38,6 +38,7 @@ public List<LocationDTO> getAllLocations() {
         }
     }
 
+    // TODO shouldn't this return a dto?
```
Daggerpov, line 41: True

Daggerpov (approved):
Everything makes sense. Made a few nit-picky comments, but I'll merge since this change is urgent.

Issue #105, opened 2025-01-22
Get Events Invited To: `FullEventDTO` needs to include friend tag colors & Participation Status


Daggerpov:
Currently, we're getting all the actual event info from @SeabertYuan 's new `FullEventDTO`, but we need to fetch which color it should display as, based on which friend tag the requesting user has assigned to the event's creator. So, I'll adjust the `FullEventDTO` to include the color hex code as a property.

Daggerpov:
Closing, since #106 merged

PR #106, opened 2025-01-22
`FullEventDTO`: including color & `ParticipationStatus` + returning from  new `invitedEvents` endpoint

Daggerpov:
#105 

Feel free to comment on anything, from naming to logic. I'm open to suggestions. 

- Rename `FullEventDTO` -> `FullFeedEventDTO` to convey purpose on mobile side of loading in feed
    - added color hex code property, as per #105 
        - Implemented its logic for retrieval in new methods: `getFriendTagColorHexCodeForRequestingUser()` & `getPertainingFriendTagByUserIds()`
- Returned `FullFeedEventDTO` from `invitedEvents` endpoint with same `full` request param as @SeabertYuan used in #100 
- Created helper method `constructDTOFromEntity(Event eventEntity)` for duplicate code fragment

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -33,7 +33,7 @@ public class EventController {
         }
     }
 
-    // full path: /api/v1/events/{id}?full=full
+    // full path: /api/v1/events/{id}?full=full&requestingUserId=requestingUserId
     @GetMapping("{id}")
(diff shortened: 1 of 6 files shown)


src/main/java/com/danielagapov/spawn/Controllers/EventController.java
```diff
@@ -181,11 +180,17 @@ public ResponseEntity<Void> toggleParticipation(@PathVariable UUID eventId, @Req
         }
     }
 
-    // full path: /api/v1/events/{userId}/invitedEvents
+    // full path: /api/v1/events/{userId}/invitedEvents?full=true
```
Daggerpov, line 183:
```suggestion
    // full path: /api/v1/events/{userId}/invitedEvents?full=full
```

PR #107, opened 2025-01-22
Implemented `getRecommendedFriendsForUserId()` + Made `RecommendedFriendUserDTO`

Daggerpov:
- Made new DTO, `RecommendedFriendUserDTO` to include a property `int mutualFriendCount`, for front-end usage in UI here (we can display "2 mutual friends" under each recommended user):
- Implemented `getRecommendedFriendsForUserId()`, which returns a list of 3 `RecommendedFriendUserDTO`s, by the most mutual friends

Open to feedback!

image

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
@@ -36,7 +36,7 @@ public interface IUserService {
     List<UserDTO> getFriendsByUserId(UUID userId);
     void removeFriend(UUID userId, UUID friendId);
     void saveFriendToUser(UUID userId, UUID friendId);
-    List<RecommendedFriendUserDTO> getRecommendedFriends(UUID id);
+    List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userIdd);
 
     // For Events:
 
--- a/src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
@@ -36,7 +36,7 @@ public interface IUserService {
     List<UserDTO> getFriendsByUserId(UUID userId);
     void removeFriend(UUID userId, UUID friendId);
     void saveFriendToUser(UUID userId, UUID friendId);
-    List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userIdd);
+    List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId);
 
     // For Events:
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -35,7 +36,7 @@ public interface IUserService {
     List<UserDTO> getFriendsByUserId(UUID userId);
     void removeFriend(UUID userId, UUID friendId);
     void saveFriendToUser(UUID userId, UUID friendId);
-    List<UserDTO> getRecommendedFriends(UUID id);
+    List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userIdd);
```
Shane-M-1, line 39: small typo: "userIdd"

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -256,12 +255,54 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+    // `RecommendedFriendUserDTO`s, to include the `mutualFriendCount`
+    public List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId) {
+        // Fetch the requesting user's friends
+        List<UUID> requestingUserFriendIds = getFriendUserIdsByUserId(userId);
+
+        // Create a set of the requesting user's friends for quick lookup
+        Set<UUID> requestingUserFriendSet = new HashSet<>(requestingUserFriendIds);
+
+        // Get all users from the repository (using getUserDTOs for detailed info)
+        List<UserDTO> allUsers = getUserDTOs();
```
Shane-M-1, line 268: This looks like it could be a potentially expensive operation

Daggerpov: Good point. Maybe I can just get the `requestingUserFriendIds` as I did above, then place all mutual friends (friends of friends) in the hashset as I did, then only search between those, finding the most common ones.

Shane-M-1 (commented):
Everything looks correct

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Shane-M-1: Yes! I think that would be a better approach -- we only need to look "1 hop" away for mutuals.

Daggerpov: Reworked in [this commit](https://github.com/Daggerpov/Spawn-App-Back-End/pull/107/commits/d31512858ccba3f9b3c4f04bfbbb67843d820c87). Let me know what you think.

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
Daggerpov: Thanks for pointing this out. Fixed; resolving

PR #109, opened 2025-01-27
AWS S3

ShaneMander:
Profile pictures are now served with S3 + Cloudfront (CDN). Chose Cloudfront just because we already using AWS for buckets so might as well just stay with it.


src/main/java/com/danielagapov/spawn/Services/S3/IS3Service.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.UUID;
+
+public interface IS3Service {
+    String putObject(byte[] file, String key);
+    void deleteObjectFromUser(UUID id);
+    String putObject(byte[] file);
+    UserDTO putObjectWithUser(byte[] file, UserDTO user);
+    UserDTO updateProfilePicture(byte[] file, UUID id);
+}
```
Daggerpov, line 14:
```suggestion
    String putObjectWithKey(byte[] file, String key);
    void deleteObjectByUserId(UUID userId);
    String putObject(byte[] file);
    UserDTO putObjectWithUser(byte[] file, UserDTO user);
    UserDTO updateProfilePicture(byte[] file, UUID userId);
}
```

Specifying which kind of id + keeping consistent with your convention of "With_" in method names, like you did in `putObjectWithUser()`. These are suggestions for naming changes; whatever you end up going with, please apply to the methods in `S3Service.java`

src/main/java/com/danielagapov/spawn/S3/S3Config.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3Client;
+
+@Configuration
+public class S3Config {
+
+    @Bean
+    public S3Client s3Client() {
```
Daggerpov, line 12: Might be better to put `S3Config` and `SecurityConfig` into a folder called `spawn/Config` instead of their individual folders, `S3` and `Security`. That seems to be a common folder structure for Java Spring Boot ([source](https://malshani-wijekoon.medium.com/spring-boot-folder-structure-best-practices-18ef78a81819#:~:text=Config%3A%20Contains%20configuration%20classes%2C%20where%20you%20configure%20application%20settings%2C%20or%20AppConfig%20for%20other%20application%2Dlevel%20configurations.))

image

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -96,9 +99,9 @@ public ResponseEntity<List<UserDTO>> getUsersByFriendTag(@PathVariable UUID tagI
 
     // full path: /api/v1/users
     @PostMapping
-    public ResponseEntity<UserDTO> createUser(@RequestBody UserDTO newUser) {
+    public ResponseEntity<UserDTO> createUser(@RequestParam("user") UserDTO newUser, @RequestParam("pfp") byte[] file) {
         try {
-            return new ResponseEntity<>(userService.saveUser(newUser), HttpStatus.CREATED);
```
Daggerpov, line 101: In here, could we handle null checking for the `pfp` request param? Such that, if I post to this endpoint without the `pfp` request param, there shouldn't be a call to `s3Service.putObjectWithUser()`. 

So, there'd be two branches:
1. What you've written, `if (file != null){}`
2. What we previously had `else`, that just calls `saveUser(newUser)`

I want to make sure that we can still call this endpoint without specifying a profile picture, since it isn't a required action to take on the UI side of things.

Daggerpov (changes requested):
Left a few suggestions/questions about your implementation

Daggerpov:
Could you adjust the `profilePicture` attribute name in `User` and `UserDTO` to something like `profilePictureUrlString`? I'll do the same on the mobile end to make this more clear.

src/main/java/com/danielagapov/spawn/Services/S3/IS3Service.java
ShaneMander: Renamed methods accordingly

src/main/java/com/danielagapov/spawn/S3/S3Config.java
ShaneMander: Made a Config package

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
ShaneMander: I addressed this [here](6ab7455ffd8743003813e2d3fde79e57f2bd68c9) but this also raises a discussion about default profile pictures and how we should serve them.

src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
```diff
@@ -47,7 +47,7 @@ public String putObject(byte[] file, String key) {
 
     public String putObject(byte[] file) {
         String key = UUID.randomUUID().toString();
-        return putObject(file, key);
+        return putObjectWithKey(file, key);
     }
 
     public UserDTO putObjectWithUser(byte[] file, UserDTO user) {
```
Daggerpov, line 53:
```suggestion
    public UserDTO putProfilePictureWithUser(byte[] file, UserDTO user) {
```
The other methods you wrote are very extensible to other types of objects than only profile pictures, but this one is highly coupled with the idea of a user's profile picture, so I think it makes more sense to name it with that in mind, rather than a name that implies it is more generic for any type of object.

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -94,7 +94,10 @@ public ResponseEntity<List<UserDTO>> getUsersByFriendTag(@PathVariable UUID tagI
     @PostMapping
     public ResponseEntity<UserDTO> createUser(@RequestParam("user") UserDTO newUser, @RequestParam("pfp") byte[] file) {
         try {
-            return new ResponseEntity<>(userService.saveUser(s3Service.putObjectWithUser(file, newUser)), HttpStatus.CREATED);
+            if (file != null) {
+                newUser = s3Service.putObjectWithUser(file, newUser);
+            }
+            return new ResponseEntity<>(userService.saveUser(newUser), HttpStatus.CREATED);
```
Daggerpov, line 100:
```suggestion
            if (file != null) {
                newUser = s3Service.putObjectWithUser(file, newUser);
            } else {
                // no file provided, so we'll assign a default profile picture file's url string
                newUser = s3Service.assignDefaultProfilePictureUrlString(newUser);
            }
            return new ResponseEntity<>(userService.saveUser(newUser), HttpStatus.CREATED);
```

src/main/java/com/danielagapov/spawn/Mappers/UserMapper.java
```diff
@@ -16,7 +16,7 @@ public static UserDTO toDTO(User user, List<UUID> friendUserIds, List<UUID> frie
                 user.getId(),
                 friendUserIds,
                 user.getUsername(),
-                user.getProfilePicture(),
+                user.getProfilePicUrlString(),
```
Daggerpov, line 19: Please change `ProfilePic` & `profilePic` occurrences to non-abbreviated (`profilePicture`) to avoid unexpected JSON parsing on the mobile end

src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -28,7 +28,7 @@ public class User implements Serializable {
         private UUID id;
         @Column(nullable = false, unique = true) // Ensures the username is unique and not null
         private String username;
-        private String profilePicture; // TODO: reconsider data type later
+        private String profilePicUrlString; // TODO: reconsider data type later
```
Daggerpov, line 31:
```suggestion
        private String profilePicUrlString;
```

Daggerpov (changes requested):
Thanks for addressing previous feedback. I like your idea about the default profile picture url string, so please create a method like I outlined in a comment under `UserController` to do that

src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
```diff
@@ -0,0 +1,108 @@
...
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import java.util.UUID;
+
+
+@Service
+public class S3Service implements IS3Service {
+    private static final String BUCKET = "spawn-pfp-store";
+    private static final String CDN_BASE = Dotenv.load().get("CDN_BASE");
```
Daggerpov, line 21: You could add a variable here
```suggestion
    private static final String CDN_BASE = Dotenv.load().get("CDN_BASE");
    private static final String DEFAULT_PROFILE_PICTURE_URL_STRING = "something";
```

src/main/java/com/danielagapov/spawn/Mappers/UserMapper.java
ShaneMander: Renamed

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
ShaneMander: I prefer the logic to take place in `putObjectWithUser` which makes only a small difference in the implementation as seen [here](d69e364cb01eefb73d5383dedebf6a8ed5968d68)

src/main/java/com/danielagapov/spawn/Models/User.java
ShaneMander: Renamed

Daggerpov (approved):
Makes sense how you null checked inside the service method itself, instead of creating two conditional branches within the endpoint. Great PR; good job!

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
Daggerpov: Yea, makes a lot of sense, and simplifies the logic here

Issue #112, opened 2025-01-28
Auth Endpoint: Determine if Google user has onboarded


Daggerpov:
Could you make an endpoint where, given google credentials, we’ll return a tuple of (true, FullUserDTO) if the user has gone through the onboarding flow? Otherwise, it’ll just return (false, null).

Basically, I can currently determine whether the user is already signed into Google through the front-end google auth setup, but I can’t determine whether to put them through the onboarding flow, or straight to the app with the FullUserDTO. So, this endpoint would help with that

PR #113, opened 2025-01-29
Auth

ShaneMander:
Implemented the endpoint described in #112. 

This solution works, but it seems unnecessary to provide a boolean along with the FullUserDTO since if the endpoint returns null, then that alone implies the user has not been onboarded.

The endpoint is currently returning the tuple of <boolean, FullUserDTO> but I think we can simplify this and refactor to just return FullUserDTO (or null)

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
@@ -10,7 +10,6 @@ import org.springframework.security.oauth2.core.user.OAuth2User;
 import org.springframework.web.bind.annotation.*;
 
 import java.net.URI;
-import java.util.Map;
 
 @RestController()
 @RequestMapping("api/v1/oauth")
@@ -36,7 +35,7 @@ public class OAuthController {
     @GetMapping("sign-in?sub=sub")
     public ResponseEntity<FullUserDTO> signIn(@RequestParam("sub") String sub) {
         try {
-            return ResponseEntity.ok().body(oauthService.isOnboardedUser(sub));
+            return ResponseEntity.ok().body(oauthService.getUserIfExistsbyExternalId(sub));
         } catch (Exception e) {
             return ResponseEntity.internalServerError().body(null);
         }
--- a/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
@@ -34,7 +34,7 @@ public class OAuthController {
 
     // full path: /api/v1/oauth/sign-in?sub=su
     @GetMapping("sign-in?sub=sub")
-    public ResponseEntity<Map<Boolean, FullUserDTO>> signIn(@RequestParam("sub") String sub) {
+    public ResponseEntity<FullUserDTO> signIn(@RequestParam("sub") String sub) {
         try {
             return ResponseEntity.ok().body(oauthService.isOnboardedUser(sub));
         } catch (Exception e) {
--- a/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
@@ -1,6 +1,7 @@
 package com.danielagapov.spawn.Controllers;
 
 import com.danielagapov.spawn.DTOs.AbstractUserDTO;
+import com.danielagapov.spawn.DTOs.FullUserDTO;
 import com.danielagapov.spawn.DTOs.UserDTO;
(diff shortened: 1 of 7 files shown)


src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -41,6 +45,17 @@ public UserDTO makeUser(UserDTO userDTO, String externId) {
         return userDTO;
     }
 
+    @Override
+    public Map<Boolean, FullUserDTO> isUser(String externId) {
+        UserIdExternalIdMap mapping = getMapping(externId);
+        if (mapping == null) {
```
Daggerpov, line 51: I agree with you that the boolean is unnecessary and we can just return either the `FullUserDTO` or null

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
```diff
@@ -30,6 +32,16 @@ public ResponseEntity<AbstractUserDTO> googleSignIn(@AuthenticationPrincipal OAu
         }
     }
 
+    //
+    @GetMapping("/sign-in?sub=sub")
```
Daggerpov, line 36:
```suggestion
    @GetMapping("sign-in?sub=sub")
```

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -41,6 +45,17 @@ public UserDTO makeUser(UserDTO userDTO, String externId) {
         return userDTO;
     }
 
+    @Override
+    public Map<Boolean, FullUserDTO> isUser(String externId) {
```
Daggerpov, line 49:
```suggestion
    public FullUserDTO getUserIfExistsByExternalId(String externalUserId) {
```
If you go with your implementation, I think this method signature is more fitting

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
```diff
@@ -30,6 +32,16 @@ public ResponseEntity<AbstractUserDTO> googleSignIn(@AuthenticationPrincipal OAu
...
+    public ResponseEntity<Map<Boolean, FullUserDTO>> signIn(@RequestParam("sub") String sub) {
```
Daggerpov, line 37:
```suggestion
    public ResponseEntity<FullUserDTO> signIn(@RequestParam("sub") String sub) {
```
If you go with your idea

Daggerpov, line 37:
```suggestion
    public ResponseEntity<Map<Boolean, FullUserDTO>> signIn(@RequestParam("externalUserId") String externalUserId) {
```

Daggerpov (changes requested):
I agree that we should go with your simplification of not returning the boolean, and just `FullUserDTO` or null. So, I left some comments about implementing that, and to do with more precise naming for methods & var names.

Daggerpov:
Reminder: also move s3Service.putProfilePictureWithUser(file, newUser); call into the UserService::saveUser() to make sure it’s consistent from every caller, similar to the friend tag logic in UserService::saveUser()

Daggerpov:
Possible improvement, though not urgent (no current use-case):
I’m wondering, do we store any indication that an externalId is from google? Once we add apple auth, we’d probably want to store another property in that repository you have called “provider” which could be an enum of those two options, to let us know which it is. Seems like a useful piece of info to store, though, I can’t think of a use-case right now, so it’s not urgent. (edited)

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
ShaneMander: Addressed in [here](327c16711495300552735169ce2e8e238b2d24cf)

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
ShaneMander: Addressed in [here](2959537cd923411da8e8f36ad29e2fc5a3601624)

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
ShaneMander: Implemented [here](https://github.com/Daggerpov/Spawn-App-Back-End/pull/113/commits/0d55bf93c1e77c24b94f462956280c39e65b8cba)

Daggerpov (approved):
Thanks for addressing everything

PR #118, opened 2025-01-29
OAuth: Documentation and var name changes

Daggerpov:
- Documentation
- Changed var names from `id` & `sub` -> `userId` or `externalUserId`
- Code change:

Before:
```
// full path: /api/v1/oauth/sign-in?sub=su
    @GetMapping("sign-in?sub=sub")
```
After:
```
// full path: /api/v1/oauth/sign-in?externalUserId=externalUserId
    @GetMapping("sign-in")
```

Pretty sure the old path was causing this 404 (should probably return 500 error, not 404):
image
image

, given that this is the code behind it:
```
        } catch (Exception e) {
            return ResponseEntity.internalServerError().body(null);
        }
```

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
+++ b/src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
@@ -29,7 +29,13 @@ public class S3Service implements IS3Service {
         this.userService = userService;
     }
 
-
+    /**
+     * This is the closest method directly to our S3Client, which puts an object to 
+     * our S3 bucket, given a key to map it to
+     * 
+     * Returns the profile picture url string where it's now hosted through a CDN
+     */
+    @Override
     public String putObjectWithKey(byte[] file, String key) {
         PutObjectRequest request = PutObjectRequest.builder()
                 .bucket(BUCKET)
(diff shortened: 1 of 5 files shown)


src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
```diff
@@ -82,13 +104,22 @@ public UserDTO updateProfilePicture(byte[] file, UUID id) {
...
+     * Delete the associated profile picture object in S3, that pertains
+     * to a given `userId`
+     */
+    @Override
+    public void deleteObjectByUserId(UUID userId) {
+        User user = UserMapper.toEntity(userService.getUserById(userId));
         String urlString = user.getProfilePictureUrlString();
         String key = extractObjectKey(urlString);
         deleteObject(key);
+        // ? shouldn't we also set the user's `profilePictureUrlString` attribute to null here?
```
Daggerpov, line 117: @ShaneMander thoughts?

ShaneMander: Yes good catch. I'll make that change later today and push it to the s3 branch

Daggerpov: Sounds good. I'll make an issue for this so we don't forget, since these little things are stacking up.

Issue #119, opened 2025-01-29
Auth: wipe `profilePictureUrlString` upon object deletion


Daggerpov:
<img width="803" alt="Image" src="https://github.com/user-attachments/assets/92eb073a-2b11-4ff1-8995-e6968aa1fcf9" />

Issue #116, opened 2025-01-29
Auth: External Id Provider Property


Daggerpov:
Possible improvement, though not urgent (no current use-case):
I’m wondering, do we store any indication that an externalId is from google? Once we add apple auth, we’d probably want to store another property in that repository you have called “provider” which could be an enum of those two options, to let us know which it is. Seems like a useful piece of info to store, though, I can’t think of a use-case right now, so it’s not urgent. (edited)

_Originally posted by @Daggerpov in https://github.com/Daggerpov/Spawn-App-Back-End/issues/113#issuecomment-2620751402_

PR #121, opened 2025-01-30
Auth issues

### LINKED ISSUE #115: Auth Duplicate-Checking
Image

ShaneMander:
Resolves #115, #116, #119

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
@@ -4,6 +4,7 @@ import com.danielagapov.spawn.DTOs.AbstractUserDTO;
 import com.danielagapov.spawn.DTOs.FullUserDTO;
 import com.danielagapov.spawn.DTOs.UserDTO;
 import com.danielagapov.spawn.Enums.OAuthProvider;
+import com.danielagapov.spawn.Helpers.Logger.ILogger;
 import com.danielagapov.spawn.Services.OAuth.IOAuthService;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.core.annotation.AuthenticationPrincipal;
@@ -16,9 +17,11 @@ import java.net.URI;
 @RequestMapping("api/v1/oauth")
 public class OAuthController {
     private final IOAuthService oauthService;
+    private final ILogger logger;
 
-    public OAuthController(IOAuthService oauthService) {
+    public OAuthController(IOAuthService oauthService, ILogger logger) {
         this.oauthService = oauthService;
+        this.logger = logger;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -31,24 +30,31 @@ public AbstractUserDTO verifyUser(OAuth2User oauthUser) {
...
 
     @Override
-    public FullUserDTO getUserIfExistsbyExternalId(String externalUserId) {
+    public FullUserDTO getUserIfExistsbyExternalId(String externalUserId, String email, OAuthProvider provider) {
         UserIdExternalIdMap mapping = getMapping(externalUserId);
-        return mapping == null ? null : getFullUserDTO(mapping);
+        if (mapping != null) {
+            return getFullUserDTO(mapping);
+        } else {
+            FullUserDTO fullUserDTO = userService.getUserByEmail(email);
```
Daggerpov, line 52: This method now has side-effects of saving the mapping, and 'does more than one thing'. The name implies you're getting something, but you might end up saving something. 

One way to fix this is instead of calling `getUserIfExistsbyExternalId()` directly from `OAuthController`, make a method called `signIn()` that then calls this, and another method if this method returns null, which then does this mapping saving.

As a software design 'rule', functions should be more pure than this, and do one thing, to be clear about their implementation. I've actually noticed while trying to understand some of your methods, that this is a common trend and makes it a little harder to follow things (subjectively, in my opinion).

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -31,24 +30,31 @@ public AbstractUserDTO verifyUser(OAuth2User oauthUser) {
...
+            if (fullUserDTO != null) {
+                saveMapping(externalUserId, fullUserDTO, provider);
```
Daggerpov, line 54: Per Slack, let's not save additional mappings to the same user anyway

src/main/java/com/danielagapov/spawn/Controllers/OAuthController.java
```diff
@@ -69,9 +70,14 @@ public ResponseEntity<FullUserDTO> signIn(@RequestParam("externalUserId") String
      */
     // full path: /api/v1/oauth/make-user
     @PostMapping("make-user")
-    public ResponseEntity<UserDTO> makeUser(@RequestParam("user") UserDTO userDTO, @RequestParam("externalUserId") String externalUserId, @RequestParam("pfp") byte[] profilePicture) {
+    public ResponseEntity<UserDTO> makeUser(
+            @RequestParam("user") UserDTO userDTO,
+            @RequestParam("externalUserId") String externalUserId,
+            @RequestParam("pfp") byte[] profilePicture,
```
Daggerpov, line 76: Make sure these are not required fields, like in my recent commits, which you can merge from `main` into your branch

Daggerpov, line 76:
```suggestion
            @RequestParam(value="profilePicture", required=false) byte[] profilePicture,
```

Daggerpov (changes requested):
Overall, great changes. Just a few things to continue discussing and revise

Daggerpov (approved):
🔥

PR #136, opened 2025-02-05
hopefully resolve duplicate issues

ShaneMander:
Made a change to `OAuthService::makeUser` to detect whether a user exists before trying to save the user. This is a temporary fix for now because I think the root of the sign-in issues is somewhere else

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
@@ -44,12 +44,17 @@ public class OAuthService implements IOAuthService {
 
     @Override
     public FullUserDTO makeUser(UserDTO userDTO, String externalUserId, byte[] profilePicture, OAuthProvider provider) {
-        // TODO: temporary solution
         try {
-            if (userExistsByExternalIdOrEmail(externalUserId, userDTO.email())) {
-                logger.log(String.format("Existing user detected in makeUser: {user: %s, externalUserId: %s}", userDTO.email(), externalUserId));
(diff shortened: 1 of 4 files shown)


src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -80,6 +85,18 @@ public FullUserDTO getUserIfExistsbyExternalId(String externalUserId, String ema
         }
     }
 
+    /**
+     * Checks if user exists, first by externalUserId then by email
+     * This is a temporary solution to duplicates occurring in database
+     */
+    private boolean userExistsByExternalIdOrEmail(String externalUserId, String email) {
```
Daggerpov, line 92: It’d be nice to know if it’s the external user id or if it’s the email that’s a duplicate, so could you take what’s in this method and put that into the saveUser() conditional? As such, you can adjust the log string to say whether it’s a duplicate external user id or email.

Daggerpov (changes requested):
Good idea to add this in, and I don’t think it needs to eventually leave, since it’s a good guard rail to have. I just suggested one change for more clear logging + you can remove the TODO comment you wrote about it being temporary.

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
ShaneMander: Changed to checking for existing mapping or existing email separately to resolve this

Daggerpov (approved):
Solid!

PR #137, opened 2025-02-06
Enhanced service test suite for ChatMessage and Event methods

evannawfal:
All old tests are now passing, and added new tests. small change made to service method getFriendTagColorHexCodeForRequestingUser in EventService to return a default hex color if null.


Daggerpov (approved):
Wow! I'm really glad you've gone ahead and done this. I'm sure @ShaneMander would appreciate it too. 

I left a suggestion for which default color to use for friend tags, but won't request that change before merging, since that case should never arise anyway (that field isn't nullable on mobile).

I skimmed through the tests and realized we might be warranted in using the Mediator pattern for our dependencies (considering just how many dependencies we have to mock), but with our back-end almost complete, it's not very important.

Daggerpov:
@evannawfal Seems like there are currently some issues with `FriendTagService.java`, or maybe with how I'm calling it from mobile. Did your test-writing uncover any bugs from that file? 

image

evannawfal:
I found one line that i fixed along with the tests pushed:
```
        if (!repository.existsById(userId)) { ...
```
to
```
        if (!userRepository.existsById(userId)) {...
```
in the saveUserToFriendTag method

PR #139, opened 2025-02-06
Syntax check

evannawfal:
To avoid accidentally pushing code that doesn't compile


Daggerpov (approved):
This is some gourmet stuff. It'd prevent a lot of Railway crashes.

PR #148, opened 2025-02-07
Syntax check

evannawfal:
Syntax checker + running tests before PR can be reviewed & before pushing to main branch.

Also fixed a few failing tests + bugfix for toggleParticipation in EventService


src/test/java/com/danielagapov/spawn/ServiceTests/EventServiceTests.java
```diff
@@ -647,24 +647,6 @@ void inviteUser_ShouldReturnTrue_WhenUserAlreadyInvited() {
         verify(eventUserRepository, never()).save(any(EventUser.class));
     }
 
-    @Test
-    void toggleParticipation_ShouldToggleStatus_WhenUserIsInvitedOrParticipating() {
```
Daggerpov, line 651: Could we fix this test instead of removing it?

Daggerpov (changes requested):
Good to approve once we fix the deleted test case

Daggerpov (approved):
Good stuff!

PR #153, opened 2025-02-09
Ordered Feed + Removing Expired Events

### LINKED ISSUE #152: Ordering feed events by date descending + not sending back ended events
Currently, from our `feedEvents` endpoint in `EventController`, we return the user's own created events, followed by the events they're invited to/participating in. I'd like to make sure that within these two sub-categories, they're ordered by which is going to happen the soonest. 

Also, we need to avoid sending back events that have already ended.

evannawfal:
Resolves #152

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
@@ -472,8 +472,8 @@ public class EventService implements IEventService {
 
         OffsetDateTime now = OffsetDateTime.now();
 
-        eventsCreated.removeIf(event -> event.getEndTime().isBefore(now));
-        eventsInvitedTo.removeIf(event -> event.getEndTime().isBefore(now));
+        eventsCreated.removeIf(event -> event.getEndTime() != null && event.getEndTime().isBefore(now));
+        eventsInvitedTo.removeIf(event -> event.getEndTime() != null && event.getEndTime().isBefore(now));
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -465,10 +466,18 @@ public List<FullFeedEventDTO> getFullEventsInvitedTo(UUID id) {
...
     public List<FullFeedEventDTO> getFeedEvents(UUID requestingUserId) {
-        List<FullFeedEventDTO> eventsCreated = convertEventsToFullFeedSelfOwnedEvents(getEventsByOwnerId(requestingUserId), requestingUserId);
+        List<FullFeedEventDTO> eventsCreated =
+                convertEventsToFullFeedSelfOwnedEvents(getEventsByOwnerId(requestingUserId), requestingUserId);
         List<FullFeedEventDTO> eventsInvitedTo = getFullEventsInvitedTo(requestingUserId);
 
-        // Combine the lists with eventsCreated first
+        OffsetDateTime now = OffsetDateTime.now();
+
+        eventsCreated.removeIf(event -> event.getEndTime().isBefore(now));
```
Daggerpov, line 475: I think we need to handle null pointer exceptions with `.getEndTime()`

Daggerpov (changes requested):
Logic seems good; just one null check missing

Daggerpov (approved):
Logic makes sense

PR #155, opened 2025-02-10
Everyone tag creation & retrieval fix

Daggerpov:
Fix involves the classic problem we've had of not using persisted ids, and instead using dto ids (which might differ) 

Problem fixed (upon clicking "accept" on friend request):

image


evannawfal (approved):
Looks good

PR #157, opened 2025-02-10
BetaAcessSignUp & OAuth Service Tests Implemented

evannawfal:
All tests passing


Daggerpov (approved):
Makes sense to emulate the front-end logic of filling the full user dto info in, by having that private helper method. Thanks for adding these tests to finish off that tests issue!

PR #158, opened 2025-02-11
Recommend friend bugfix

evannawfal:
Fixed 404 Error


Daggerpov (approved):
Thought of the same thing, so if we both did, it must be right, right?

PR #159, opened 2025-02-11
Jwt only

ShaneMander:
JWT only


Daggerpov (changes requested):
Waiting for build checks to pass, since we don't want to merge this in and have it depend on the other changes in #132 . Ideally, we split them to where this one can be merged alone and still work

ShaneMander:
@Daggerpov Breaking changes: the `oauth` endpoint no longer exists. It has been renamed to [auth](url) so email + pass and oauth sign-in can be accessed from the same controller

ShaneMander:
@Daggerpov can we get this merged?

Daggerpov (approved):
Adjusted endpoints in build 53 to work with this PR

Issue #154, opened 2025-02-10
Accepting/Rejecting Friend Request Bugfix


evannawfal:
Currently, when a user attempts to accept or reject a friend request, no visible changes occur in our app. The status of the friend request remains unchanged in the DB as well. Source of bug needs to be found + fixed :).

Daggerpov:
Btw, the incoming friend requests gets re-fetched from iOS when navigating between friends<->tags. So, I'm pretty sure this is a back-end issue.

Daggerpov:
Once we can confirm this is fixed with your PR by manually testing through the app, let's resolve this @evannawfal

Daggerpov:
Should be good

PR #161, opened 2025-02-12
Reworked friend req endpoints

Daggerpov:
- more consise endpoints
    - consolidated accept & reject endpoints into one with a new `@RequestParam` of new enum type `FriendRequestAction`
    - removed unused `userId` path variable arguments from them
    - removed prefix `/users/` in URL
- try-catching in `acceptFriendRequest()` (cherry-picked this commit into `main` already)

To note, this does not address the bug in #154


evannawfal (approved):
Nice fix

Issue #131, opened 2025-02-04
Friend Search Filtering


Daggerpov:
- [x] Make an endpoint with a `String` search query for friends, which will filter down the `UUID requestingUserId`'s list of friends, to return as `List<RecommendedFriendUserDTO>`
    - [x] Should be under `UserController`
    - [x] path could be something like `/api/v1/users/getFilteredFriends/{requestingUserId}`
    - [x] `@PathVariable UUID requestingUserId`
    - [x] `@RequestParam String searchQuery`
- [x] Remember to populate the `mutualFriendCount` propery of this DTO
- [x] First start by making this work, then potentially explore a way to make performant. 

From @SeabertYuan and I on Slack:

<img width="946" alt="Image" src="https://github.com/user-attachments/assets/526b41c8-bd2a-40fd-86fc-091e460aa0f5" />

<img width="339" alt="Image" src="https://github.com/user-attachments/assets/c71bc9c1-4685-45d6-acc4-dc972fec92fd" />

SeabertYuan:
Going based on the REST API convention shared, I think the full path should be something more like `/api/v1/users/{requestingUserId}?query=searchQuery`.

Daggerpov:
> Going based on the REST API convention shared, I think the full path should be something more like `/api/v1/users/{requestingUserId}?query=searchQuery`.

Yea, probably a combination of our two ideas: `/api/v1/users/getFilteredFriends/{requestingUserId}?searchQuery`. I feel like the `/users/` endpoint is too general, and this way we're more explicit that its to get filtered friends, even if that differs from REST convention (which we do pretty often with using verbs, which I think fall under RPCs more than REST endpoints)

PR #171, opened 2025-02-13
Friend search filtering

SeabertYuan:
Addresses #131. ***NOT complete yet.***

- [ ] Convert `List<User>` into `List<RecommendedFriendUserDTO>`.

I've been out of the loop for too long so I have to study how our DTOs work now after a bunch of changes that @Daggerpov made.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/UserController.java
@@ -173,6 +173,12 @@ public class UserController {
         }
     }
 
+    // full path: /api/v1/users/{requestingUserId}?query=searchQuery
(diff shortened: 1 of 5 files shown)


Daggerpov:
For converting users -> RecommendedFriendUserDTO it first involves calling getFullUserByUser(), and from FullUserDTO -> RecommendedFriendUserDTO it just adds one property, mutualFriendsCount which is calculated in getRecommendedFriends(). The problem is, that calculation is coupled inside that function. So, the first step would be to split up that getRecommendedFriends() function into just the algorithm + other parts, such that you can use the algorithm for this PR.

SeabertYuan:
Too many merge conflicts.

Issue #172, opened 2025-02-13
(App Store Compliance) User Account Deletion Endpoint


Daggerpov:
We need to make it so that our DELETE user endpoint deletes all of its related data, including:
- external id mapping
- events it’s invited to, participating, and has created
- friend tags
- friend tags it’s apart of
- friend requests it’s involved in

ShaneMander:
With this, I think we should be able to leverage our database constraints -- we can adjust our schemas so that any foreign key referencing a user has the ON DELETE CASCADE constraint.

PR #177, opened 2025-02-16
User Delete Account Endpoint

ShaneMander:
For #172.

Added the `@OnDelete` annotation to models with a foreign key reference to a User entity so that deleting a user cascades to child tables. Since our database is already initialized, this annotation alone does not affect our database, however, it is useful to communicate our intent between tables and also so that if we create a new database, it will be correctly initialized.

I have manually added the cascading delete constraint through MySQL to each of the relevant tables in our database. So now, deleting a user will automatically delete their events, chat messages (+ likes), friend tags, friend requests, etc. 

I have also added `@OnDelete` annotations to other tables. For example, deleting an event should also delete any attached chat messages. Similarly, deleting a chat message should also delete any attached likes. I have added the cascading delete constraint to these tables as well.

Made deleting a profile picture transactional -- previously, in the case that a request to delete a user fails (i.e. the user does not end up deleted) that user's profile picture was deleted anyway. Now, a user's profile is deleted only if the user was successfully deleted.

Updated tests that invoked `UserService::deleteById` so they pass (were failing due to the above change).


src/main/java/com/danielagapov/spawn/Services/S3/S3Service.java
```diff
@@ -143,15 +142,29 @@ public String getDefaultProfilePicture() {
...
+            throw e;
+        }
+    }
+
     /**
      * Deletes an object given the key (where it's stored)
      */
     private void deleteObject(String key) {
+        if (key == null || key.equals(extractObjectKey(DEFAULT_PFP)))
+            return; // Don't delete the default pfp! It is shared among many users
```
Daggerpov, line 162: smarat

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -241,12 +241,10 @@ public UserDTO replaceUser(UserDTO newUser, UUID id) {
...
     @Override
     public boolean deleteUserById(UUID id) {
-        if (!repository.existsById(id)) {
-            throw new BaseNotFoundException(EntityType.User, id);
-        }
-
+        User user = repository.findById(id).orElseThrow(() -> new BaseNotFoundException(EntityType.User, id));
         try {
             repository.deleteById(id);
+            s3Service.deleteObjectByURL(user.getProfilePictureUrlString());
```
Daggerpov, line 247: makes sense to do it this way like you put in description, only deleting pfp if user got deleted

Daggerpov (approved):
Great work!

PR #190, opened 2025-02-20
(HOLDING OFF MERGE) Adjust make user endpoint for images

Daggerpov:
Was having a LOT of difficulty passing the profile picture data through a request param on mobile, so I've created a new DTO, `UserCreationDTO` to contain the profile picture raw byte data. Then, I'll pass this through a new wrapper method `createUser()` to simply get this profile picture data and call the existing `makeUser()` method as before. 

The whole point of this is to avoid sending the image through a request param, but through the DTO. 

image

image


src/main/java/com/danielagapov/spawn/DTOs/UserCreationDTO.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.UUID;
+
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class UserCreationDTO implements Serializable, IOnboardedUserDTO {
```
ShaneMander, line 16: Should implement AbstractUserDTO but not IOnboardedUserDTO, right?

ShaneMander (changes requested):
Ah, passing the image through the request body via a new DTO actually makes more sense, nice.

src/main/java/com/danielagapov/spawn/DTOs/UserCreationDTO.java
Daggerpov: Yup, good catch

ShaneMander (approved):

Issue #180, opened 2025-02-16
Apple OAuth: handle external user id in `sign-in` method, without email


Daggerpov:
This is what apple external user id looks like: 000897.eede6b50e2a9498d821a236dabeddffe.0226

From apple, email isn't guaranteed, but the endpoint should still be able to verify if a user account exists by external user id (from apple). 

So, I think we need to make this request param not required. 

If provided -> do old flow for google
if not -> do slightly adapted flow for apple

Daggerpov:
@ShaneMander done right?

PR #191, opened 2025-02-20
change requirements for make-user endpoint

ShaneMander:
For #180.

For apple oauth, the only necessary change is to remove the requirement for email in the sign-in endpoint and only try checking if a user exists by email, when email is not null.

For the make-user endpoint, there is no longer any reason for the `externalId ` argument to not be required, since email+pass authentication will be done at a separate endpoint. Additionally, there is no reason for the `provider` argument to not be required; it should always be one of "google" or "apple". This will be a breaking change if not already provided from Mobile.


Daggerpov:
Provider being required is fine on the mobile side. Thanks for pointing that out though!

Daggerpov (approved):
Nice attention to detail with the null checks in `OAuthService`

PR #192, opened 2025-02-20
Beta Access Sign Up Property Adjustments

Daggerpov:
Added `hasSubscribedToNewsletter` + removed first & last name, Instagram, and additional comments

### DIFF
--- a/src/main/java/com/danielagapov/spawn/DTOs/BetaAccessSignUpDTO.java
+++ b/src/main/java/com/danielagapov/spawn/DTOs/BetaAccessSignUpDTO.java
@@ -16,9 +16,6 @@ import java.util.UUID;
 public class BetaAccessSignUpDTO implements Serializable {
     private UUID id;
     private String email;
-    private String firstName;
-    private String lastName;
     private OffsetDateTime signedUpAt;
-    private String additionalComments;
-    private String instagramUsername;
+    private Boolean hasSubscribedToNewsletter;
 }
--- a/src/main/java/com/danielagapov/spawn/Models/BetaAccessSignUp.java
+++ b/src/main/java/com/danielagapov/spawn/Models/BetaAccessSignUp.java
@@ -26,11 +26,8 @@ public class BetaAccessSignUp implements Serializable {
     @GeneratedValue(strategy = GenerationType.AUTO)
     private UUID id;
     private String email;
-    private String firstName;
-    private String lastName;
     private OffsetDateTime signedUpAt;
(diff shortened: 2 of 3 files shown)


src/main/java/com/danielagapov/spawn/DTOs/BetaAccessSignUpDTO.java
```diff
@@ -16,9 +16,6 @@
 public class BetaAccessSignUpDTO implements Serializable {
     private UUID id;
     private String email;
-    private String firstName;
-    private String lastName;
     private OffsetDateTime signedUpAt;
```
ShaneMander, line 19: What are we planning on using sign up time for?

src/main/java/com/danielagapov/spawn/Models/BetaAccessSignUp.java
```diff
@@ -26,11 +26,8 @@ public class BetaAccessSignUp implements Serializable {
     @GeneratedValue(strategy = GenerationType.AUTO)
     private UUID id;
     private String email;
```
ShaneMander, line 28: Should we make this unique + not null?

ShaneMander (changes requested):
Everything looks good, just a couple questions

src/main/java/com/danielagapov/spawn/DTOs/BetaAccessSignUpDTO.java
Daggerpov: Useful info, mainly for safe-keeping.

src/main/java/com/danielagapov/spawn/Models/BetaAccessSignUp.java
Daggerpov: Great points! I've adjusted those accordingly.

src/main/java/com/danielagapov/spawn/DTOs/BetaAccessSignUpDTO.java
Daggerpov: If we want to send out batches of invites (say, first 20, then next 20, this is useful for querying based on datetime)

Daggerpov: Having more things in our DTO isn't ever a hinderance, since we can just ignore those properties in the JSON DTOs we send to our api (POST requests); I'd rather have more attributes than less.

PR #205, opened 2025-02-22
Friend search filtering new

SeabertYuan:
implements very basic search functionality to address #131.

**Important**:
- This function is not finished for recommended friends so the search will not do any searching for "new users."

Next steps:
- [x] Refactor `getRecommendedFriendsForUserId()` into some helper functions which can be reused for the search query comparison logic
- [x] Test the code so we @Daggerpov does not have a headache later
- [x] Wrap in try-catch blocks for more robust error handling
- [ ] ~~Implement more advanced search feature.~~


SeabertYuan:
There's a problem with using the Java `Triple` since it's from an external library `org.apache.common.lang3`. Any ideas on how we want to proceed?

Daggerpov:
> There's a problem with using the Java `Triple` since it's from an external library `org.apache.common.lang3`. Any ideas on how we want to proceed?
> 
> 1. I think if we use `org.antlr.v4.runtime.misc` it should be ok, Maybe it's not a good idea to be importing random things though.
> 2. We can pop `org.apache.common.lang3` into our production.
> 3. I can just write a triple class and put it in a new "utils" folder or something. I think this might be the best solution because we have less dependence on external libraries.

One solution that I like is to make a class that contains these 3 lists as properties. That way, it’s our own data structure that we can call something like ‘searchedUserResults’

SeabertYuan:
I have not written tests yet, but I think this PR is pretty much good to go as far as connecting backend with frontend goes. I will try to get a basic test suite done in the next hour or so so we can merge with confidence.

One thing to note:
- Should I match the searchQuery with friend requests as well? I'm not sure how our design is supposed to work here.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
-
-                        // Add to excluded list to prevent duplicates
-                        excludedUserIds.add(potentialFriendId);
-                    }
+    @Override
+    public SearchedUserResult getRecommendedFriendsBySearch(UUID requestingUserId, String searchQuery) {
+        try {
+
+            // Step 1. Find all incoming friend Requests
+            List<FullFriendRequestDTO> incomingFriendRequests = friendRequestService.getIncomingFriendRequestsByUserId(requestingUserId);
```
Daggerpov, line 569: Great work with the search filtering for recommended & current friends. How come we're not filtering the incoming friend requests as well? I feel like, for now, you could re-use your same 'easy' logic of comparing user, first, and last name, right?

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+        }
+    }
 
-            // Create a set of the requesting user's friends, users they've sent requests to, users they've received requests from, and self for quick lookup
-            Set<UUID> excludedUserIds = new HashSet<>(requestingUserFriendIds);
-            excludedUserIds.addAll(sentFriendRequestReceiverUserIds);
-            excludedUserIds.addAll(receivedFriendRequestSenderUserIds);
-            excludedUserIds.addAll(existingFriendUserIds);
-            excludedUserIds.add(userId); // Exclude self
+    private List<RecommendedFriendUserDTO> getRecommendedMutuals(UUID userId, Set<UUID> excludedUserIds) {
```
Daggerpov, line 435: Nice refactoring this into a separate method!

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
     @Override
     public List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId) {
         try {
-            // Fetch the requesting user's friends
-            List<UUID> requestingUserFriendIds = getFriendUserIdsByUserId(userId);
+            Set<UUID> excludedUserIds = getExcludedUserIds(userId);
```
Daggerpov, line 418: Please move this call of `getExcludedUserIds` into the functions you're passing it into, instead of having to pass it as an argument. This will reduce logic coupling from the outside function callers -> function, by just having it within the functions themselves. So, from what I can see, you could move this line `Set<UUID> excludedUserIds = getExcludedUserIds(userId);` directly into these methods: `getRecommendedMutuals()` and `getRandomNRecommendations()` since that logic 'belongs' there. 

Totally subjective, but I hope you agree with me that it'd make this a tad-bit cleaner.

src/main/java/com/danielagapov/spawn/Utils/Triple.java
```diff
@@ -0,0 +1,23 @@
+package com.danielagapov.spawn.Utils;
+
+import lombok.AllArgsConstructor;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+@AllArgsConstructor
+// Represents a tuple of 3 items
+// TODO: overriding equals and hashcode might be useful
+public class Triple<X, Y, Z> {
```
Daggerpov, line 10: Nice! I think this class has all almost all it needs (+ I like the method names), but agree that you should override equals and hashcode now, to avoid future problems. I found that with the composite keys, that was actually the root of a problem, so please change it now so we don't forget later.

src/main/java/com/danielagapov/spawn/Utils/SearchedUserResult.java
```diff
@@ -0,0 +1,14 @@
+package com.danielagapov.spawn.Utils;
+
+import com.danielagapov.spawn.DTOs.FriendRequest.FullFriendRequestDTO;
+import com.danielagapov.spawn.DTOs.User.FullFriendUserDTO;
+import com.danielagapov.spawn.DTOs.User.RecommendedFriendUserDTO;
+import lombok.AllArgsConstructor;
+
+import java.util.List;
+
+public class SearchedUserResult extends Triple<List<FullFriendRequestDTO>, List<RecommendedFriendUserDTO>, List<FullFriendUserDTO>> {
```
Daggerpov, line 10: I like this a lot.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
 
-            // Fetch users who have already received a friend request from the user
-            List<UUID> sentFriendRequestReceiverUserIds = friendRequestService.getSentFriendRequestsByUserId(userId)
-                    .stream()
-                    .map(FriendRequestDTO::getReceiverUserId)
-                    .toList();
+            List<RecommendedFriendUserDTO> recommendedFriends = getRecommendedMutuals(userId, excludedUserIds).stream().limit(3).toList();
```
Daggerpov, line 420: Let's move this logic of limiting to 3 into `getRecommendedMutuals`, and have a constant var in this `UserService` class called something like "recommended friends limit" so we can easily change in future. Since, we might want to change on mobile to fetching all recommended friends at once, then controlling the number on the mobile side. 

Having the 3 here seems like such a 'magic number'.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
 
-            // Map mutual friends to RecommendedFriendUserDTO
-            List<UUID> receivedFriendRequestSenderUserIds = friendRequestService.getIncomingFriendRequestsByUserId(userId)
-                    .stream()
-                    .map(request -> request.getSenderUser().getId())
-                    .toList();
+            if (recommendedFriends.size() >= 3) {
```
Daggerpov, line 422: Replace magic number with var, as stated above. 

I know I was the one to previously implement this, so it's my bad, but since you're already improving this algorithm quite a bit, let's make it as clean as we can.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+                return recommendedFriends;
+            }
 
-            List<UUID> existingFriendUserIds = getFriendUserIdsByUserId(userId);
+            // Otherwise fill with random friends
+            recommendedFriends = getRandomNRecommendations(3 - recommendedFriends.size(), excludedUserIds, userId);
```
Daggerpov, line 427: magic number

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+                            fullUser.getEmail(),
+                            mutualFriendCount
+                    );
+                })
+                .sorted(Comparator.comparingInt(RecommendedFriendUserDTO::getMutualFriendCount).reversed())
+                .collect(Collectors.toList());
+    }
+
+    // Gets random N users for userId to be friends, if n > number of possible recommended friends, return all possible recommended friends
+    private List<RecommendedFriendUserDTO> getRandomNRecommendations(int n, Set<UUID> excludedUserIds, UUID userId) {
```
Daggerpov, line 467: Love this new method.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -414,122 +415,184 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+            boolean isExcluded = excludedUserIds.contains(potentialFriendId);
+
+            if (!isExcluded) {
+                boolean hasAlreadySentFriendRequest = false;
+
+                try {
+                    List<FullFriendRequestDTO> potentialFriendIncomingFriendRequests = friendRequestService.getIncomingFriendRequestsByUserId(potentialFriendId);
+
+                    for (FullFriendRequestDTO friendRequestDTO : potentialFriendIncomingFriendRequests) {
+                        if ((friendRequestDTO.getSenderUser().getId().equals(userId) && friendRequestDTO.getReceiverUser().getId().equals(potentialFriendId)) ||
```
Daggerpov, line 483: Could you move this if condition into a separate var, and give it a good name? Again, I know I was the one to originally write this bad code, but let's make it better now. 

Takes a few seconds to understand what it's checking for, which we want to make as easy as possible.

Daggerpov: Similar to our check below of `if (!hasAlreadySentFriendRequest)`, let's create a var here

Daggerpov (approved):
I really love the renovations you've done with the place! In general, your logic seems pretty sound; I mainly left more code cleanliness comments to improve it even further (after realizing how bad my initial algorithm's code really was lol). 

Your refactoring into methods is all really well done!

Daggerpov:
> implements very basic search functionality to address #131.
> 
> **Important**:
> 
> * This function is not finished for recommended friends so the search will not do any searching for "new users."
> 
> Next steps:
> 
> * [x]  Refactor `getRecommendedFriendsForUserId()` into some helper functions which can be reused for the search query comparison logic
> * [ ]  Test the code so we @Daggerpov does not have a headache later
> * [x]  Wrap in try-catch blocks for more robust error handling
> * [ ]  Implement more advanced search feature.

Let's implement more advanced search feature in a separate PR, and maybe leave @OnionFastener to do that. 

As for testing, I think @ShaneMander would get a worse headache than I would from missing it lol; he was the one pushing for us to test all of our code.

Daggerpov (changes requested):
Meant to click "request changes," not "approve" on this PR; please address my comments above before we merge this; don't worry about the advanced search algo for now. 

Please feel free to resolve my comments as you work through them, but if you choose not to do something because you disagree, please leave it open to discussion.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
SeabertYuan: Yeah this was the comment I left above I'm not sure what we wanted from a design stand point.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
SeabertYuan: Definitely, passing things around like that does get confusing :)

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Let's do it to keep consistent, and in case someone genuinely has like 50 incoming requests

SeabertYuan: I thought about it and I think regardless of how we intended to design it it's best if we filter it out.

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -57,4 +59,7 @@ public interface IUserService {
     boolean existsByEmail(String email);
     FullUserDTO getFullUserByUsername(String username);
     boolean existsByUsername(String username);
+
+    // TODO figure out which section this goes in
```
Daggerpov, line 63:
```suggestion
```

src/main/java/com/danielagapov/spawn/Repositories/IUserRepository.java
```diff
@@ -4,6 +4,7 @@
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.util.List;
```
Daggerpov, line 7:
```suggestion

```

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -464,9 +470,10 @@ private List<RecommendedFriendUserDTO> getRecommendedMutuals(UUID userId, Set<UU
     }
 
     // Gets random N users for userId to be friends, if n > number of possible recommended friends, return all possible recommended friends
-    private List<RecommendedFriendUserDTO> getRandomNRecommendations(int n, Set<UUID> excludedUserIds, UUID userId) {
+    public List<RecommendedFriendUserDTO> getRandomNRecommendations(Long n, UUID userId) {
```
Daggerpov, line 473: Please revert to `private`, if they don't need to be public (e.g. if they need to be accessed via the interface of `IUserService.java` or tested via this interface. They can just be unit tested through the concrete `UserService` class.

Daggerpov (changes requested):
Hey @SeabertYuan , overall, great work man. Since this PR isn't time-sensitive, I've left some more comments based on your revisions (P.S. thanks for addressing everything I wrote earlier!). 

Sorry, should be the last review. I think there were a few patterns:
- `public` methods when unnecessary
- passing the internal `recommendedFriendsLimit` var around internally, instead of simply accessing internally
    - thereby removing the need for wrappers that pass it in as an arg (in general, reduce args when possible)
- a few performance/readability little improvements

src/test/java/com/danielagapov/spawn/ServiceTests/UserServiceTests.java
```diff
@@ -266,4 +271,44 @@ void getAllUsers_ShouldReturnEmptyList_WhenNoUsersExist() {
...
\ No newline at end of file
+
+    @Test
+    void getRecommendedFriendsBySearch_ShouldWorkWithQueryFullRecommendations() {
+        UserService spyUserService = spy(userService);
+        UUID user1Id = UUID.randomUUID();
+        UUID user2Id = UUID.randomUUID();
+        UUID user3Id = UUID.randomUUID();
+        UUID user4Id = UUID.randomUUID();
+        FullUserDTO user1Full = new FullUserDTO(user1Id, List.of(), "john_doe", "profile.jpg", "John", "Doe", "A bio", List.of(), "john.doe@example.com");
```
ShaneMander, line 282: I might be missing something, but it seems like this isn't used elsewhere in the test?

src/test/java/com/danielagapov/spawn/ServiceTests/UserServiceTests.java
```diff
@@ -266,4 +271,44 @@ void getAllUsers_ShouldReturnEmptyList_WhenNoUsersExist() {
...
+        assertEquals(new SearchedUserResult(List.of(), List.of(user3Full), List.of()), res);
+    }
+    @Test
+    void getRecommendedFriendsBySearch_ShouldWorkWithQueryFullRecommendationsAndFriends() {
+        UserService spyUserService = spy(userService);
+        UUID user1Id = UUID.randomUUID();
+        UUID user2Id = UUID.randomUUID();
+        UUID user3Id = UUID.randomUUID();
+        UUID user4Id = UUID.randomUUID();
+        FullUserDTO user1Full = new FullUserDTO(user1Id, List.of(), "john_doe", "profile.jpg", "John", "Doe", "A bio", List.of(), "john.doe@example.com");
```
ShaneMander, line 299: Same as above

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
SeabertYuan: I made these public because I needed to access some of them for testing purposes and I figured it would be weird to have some of them private just because I hadn't written tests for them yet.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -549,24 +558,31 @@ private Set<UUID> getExcludedUserIds (UUID userId) {
...
         return excludedUserIds;
     }
 
     @Override
     public SearchedUserResult getRecommendedFriendsBySearch(UUID requestingUserId, String searchQuery) {
-        try {
+        return getRecommendedFriendsBySearch(requestingUserId, searchQuery, recommendedFriendsLimit);
+    }
 
+    public SearchedUserResult getRecommendedFriendsBySearch(UUID requestingUserId, String searchQuery, Long limit) {
```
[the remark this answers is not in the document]
SeabertYuan: I wanted to do this because I figured it would be more flexible if we overloaded the function to allow it to have any N as a limit. I think it's better to have this now rather than having to update the constant value and change that value.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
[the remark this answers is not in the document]
SeabertYuan: Long is because `.limit()` takes in a long value. I don't remember the exact reason but I know I thought that it was kind of weird to be casting things around like that.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -410,32 +411,37 @@ public void removeFriend(UUID userId, UUID friendId) {
...
             logger.log(e.getMessage());
             throw e;
         }
     }
 
-    private List<RecommendedFriendUserDTO> getRecommendedMutuals(UUID userId, Set<UUID> excludedUserIds) {
+    // returns top recommended friends as specified by the class constant (3)
+    @Override
+    public List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId) {
+        return getRecommendedFriendsForUserId(userId, recommendedFriendsLimit);
```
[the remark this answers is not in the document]
SeabertYuan: Yeah, I figured overloading the method would give the most flexibility. We can "default" to the constant or choose to have the limit more or less than the default.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -549,24 +558,31 @@ private Set<UUID> getExcludedUserIds (UUID userId) {
                 .map(request -> request.getSenderUser().getId())
                 .toList();
 
-        List<UUID> existingFriendUserIds = getFriendUserIdsByUserId(userId);
-
         // Create a set of the requesting user's friends, users they've sent requests to, users they've received requests from, and self for quick lookup
         Set<UUID> excludedUserIds = new HashSet<>(requestingUserFriendIds);
         excludedUserIds.addAll(sentFriendRequestReceiverUserIds);
         excludedUserIds.addAll(receivedFriendRequestSenderUserIds);
-        excludedUserIds.addAll(existingFriendUserIds);
```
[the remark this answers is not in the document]
SeabertYuan: No because requestingUserFriendIds is the same thing as existingFriendUserIds I think so we were adding the same thing twice I believe.

src/test/java/com/danielagapov/spawn/ServiceTests/UserServiceTests.java
SeabertYuan: yeah you're right I forgot to remove it lol.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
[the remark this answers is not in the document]
Daggerpov: Ah, yea, I think so too. Resolving

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: From Slack DM:

"About testing, while marking functions public makes them testable from the interface, since you intend to unit test them (aka they won’t be externally accessed from other classes via the interface), I think you should test the class itself and it’s private methods, not through the interface.

And btw, it is a great idea to test those since they’re probably the most logic heavy methods in our entire app"

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -549,24 +558,31 @@ private Set<UUID> getExcludedUserIds (UUID userId) {
...
         excludedUserIds.add(userId); // Exclude self
 
         return excludedUserIds;
     }
 
     @Override
     public SearchedUserResult getRecommendedFriendsBySearch(UUID requestingUserId, String searchQuery) {
-        try {
+        return getRecommendedFriendsBySearch(requestingUserId, searchQuery, recommendedFriendsLimit);
```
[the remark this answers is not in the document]
Daggerpov: From Slack DM:

"Yea, I see where you’re coming from, if we specify how many we want from the endpoints. I think right now though it looks strange and useless to pass an internal private var, and you might be over-abstracting for the time being"

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -409,127 +411,208 @@ public void removeFriend(UUID userId, UUID friendId) {
...
-            excludedUserIds.addAll(existingFriendUserIds);
-            excludedUserIds.add(userId); // Exclude self
+    // returns top recommended friends as specified by the class constant (3)
+    @Override
+    public List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId) {
+        return getRecommendedFriendsForUserId(userId, recommendedFriendsLimit);
+    }
+
+    public List<RecommendedFriendUserDTO> getRecommendedMutuals(UUID userId) {
+        // Fetch the requesting user's friends
```
ShaneMander, line 440: Could you add a logging statement here as well to make debugging a bit easier later? Something simple like `logger.log("Getting recommended mutuals for user id: <id>")`

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -409,127 +411,208 @@ public void removeFriend(UUID userId, UUID friendId) {
...
             }
+        }
+        return mutualFriendCounts;
+    }
 
-            // Otherwise, recommend random users not already friends, not sent/received requests, and not self
-            List<UserDTO> allUsers = getAllUsers();
+    // Create a set of the requesting user's friends, users they've sent requests to, users they've received requests from, and self for quick lookup
+    public Set<UUID> getExcludedUserIds (UUID userId) {
+        // Fetch the requesting user's friends
```
ShaneMander, line 546: Another logging statement here would be useful

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -409,127 +411,208 @@ public void removeFriend(UUID userId, UUID friendId) {
...
-                                fullUserDTO.getBio(),
-                                fullUserDTO.getFriendTags(),
-                                fullUserDTO.getEmail(),
-                                0 // No mutual friends
-                        ));
-
-                        // Add to excluded list to prevent duplicates
-                        excludedUserIds.add(potentialFriendId);
-                    }
+    public SearchedUserResult getRecommendedFriendsBySearch(UUID requestingUserId, String searchQuery, Long limit) {
```
ShaneMander, line 575: Logging statement at the start of this method can be useful

ShaneMander (dismissed):

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -409,127 +411,208 @@ public void removeFriend(UUID userId, UUID friendId) {
...
+                // Step 2. Get recommended friends
+                recommendedFriends = getRecommendedFriendsForUserId(requestingUserId);
+                // Step 3. Get all friends
+                friends = getFullFriendUsersByUserId(requestingUserId);
+            } else { // If searchQuery is not empty:
+                // Step 2. List all recommended friends who match based on searchQuery
+                // Map mutual friends to RecommendedFriendUserDTO
+                recommendedFriends = getRecommendedMutuals(requestingUserId).stream()
+                        .filter(entry -> entry.getFirstName().equals(searchQuery) ||
+                                entry.getLastName().equals(searchQuery) ||
```
OnionFastener, line 600: Could use .contains() so we can search substrings (ex search "tim" for "timmy")?

src/main/java/com/danielagapov/spawn/DTOs/User/AbstractUserDTO.java
```diff
@@ -18,4 +18,9 @@ public abstract class AbstractUserDTO implements Serializable {
     private String email;
     private String username;
     private String bio;
+
+    @Override
+    public boolean equals(Object obj) {
+        return getId().equals(((AbstractUserDTO) obj).getId());
+    }
```
Daggerpov, line 25: These `equals()` methods should be more robust to check for nulls, since that would otherwise throw a null pointer exception. 

```
@Override
public boolean equals(Object obj) {
    if (this == obj) return true;
    if (obj == null || getClass() != obj.getClass()) return false;
    AbstractUserDTO that = (AbstractUserDTO) obj;
    return id != null && id.equals(that.id);
}
```

src/main/java/com/danielagapov/spawn/Util/SearchedUserResult.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+
+// A special instance of Triple
+public class SearchedUserResult extends Triple<List<FetchFriendRequestDTO>, List<RecommendedFriendUserDTO>, List<FullFriendUserDTO>> {
+    public SearchedUserResult(List<FetchFriendRequestDTO> friendRequestDTOS, List<RecommendedFriendUserDTO> recommendedFriendUserDTOS, List<FullFriendUserDTO> fullFriendUserDTOS) {
+        super(friendRequestDTOS, recommendedFriendUserDTOS, fullFriendUserDTOS);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Daggerpov, line 16: Again, not a safe `equals()` method, as stated above

src/main/java/com/danielagapov/spawn/DTOs/FriendRequest/AbstractFriendRequestDTO.java
```diff
@@ -13,4 +13,10 @@
 @Setter
 public abstract class AbstractFriendRequestDTO {
     private UUID id;
+
+    @Override
```
Daggerpov, line 17: unsafe (doesn't handle nulls)

Daggerpov (changes requested):
Unsafe `equals`() methods since they don't handle null pointer exceptions and they don't check if the two objects are even of the same class. Was there a reason for not checking the class? Pretty standard from what we learned in Java at UBC.

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -393,118 +398,185 @@ public void saveFriendToUser(UUID userId, UUID friendId) {
...
-            // Fetch the requesting user's friends
-            List<UUID> requestingUserFriendIds = getFriendUserIdsByUserId(userId);
+            List<RecommendedFriendUserDTO> recommendedFriends = getRecommendedMutuals(userId).stream().limit(recommendedFriendLimit).toList();
 
-            // Fetch users who have already received a friend request from the user
-            List<UUID> sentFriendRequestReceiverUserIds = friendRequestService.getSentFriendRequestsByUserId(userId)
-                    .stream()
-                    .map(CreateFriendRequestDTO::getReceiverUserId)
-                    .toList();
+            if (recommendedFriends.size() >= recommendedFriendLimit) {
```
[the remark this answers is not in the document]
Daggerpov: Also, if we essentially did get too many recommended friends, why would we then return them all?

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -393,118 +398,185 @@ public void saveFriendToUser(UUID userId, UUID friendId) {
...
+            boolean isExcluded = excludedUserIds.contains(potentialFriendId);
+
+            if (!isExcluded) {
+                boolean hasAlreadySentFriendRequest = false;
+
+                try {
+                    List<FetchFriendRequestDTO> potentialFriendIncomingFriendRequests = friendRequestService.getIncomingFetchFriendRequestsByUserId(potentialFriendId);
+
+                    for (FetchFriendRequestDTO friendRequestDTO : potentialFriendIncomingFriendRequests) {
+                        boolean potentialFriendSentOrReceived = friendRequestDTO.getSenderUser().getId().equals(userId);
```
Daggerpov, line 481: this var name `potentialFriendSentOrReceived` is misleading. It only checks if the potentialFriendRequest was sent. Could this be a logic issue?

Daggerpov: Matter of fact, why are we even checking for these friend requests in the first place? Wouldn't those have already been computed with the method `getExcludedUserIds()`?

Daggerpov (changes requested):
Redundant friend request checking logic

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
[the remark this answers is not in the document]
Daggerpov: Addressed. Resolving

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Along that same thought, we should just be able to check below if the potential friend user has already been excluded

Daggerpov: Addressed myself. Resolving

src/main/java/com/danielagapov/spawn/Util/SearchedUserResult.java
Daggerpov: Addressed myself. Resolving.

src/main/java/com/danielagapov/spawn/DTOs/User/AbstractUserDTO.java
Daggerpov: Addressed myself. Resolving.

Daggerpov (approved):

PR #208, opened 2025-02-23
Change Custom Header name

ShaneMander:
I had a space in a custom header name ("Refresh Token") that I think might have been causing issues.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/AuthController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/AuthController.java
@@ -105,7 +105,7 @@ public class AuthController {
     private HttpHeaders makeHeadersForTokens(String username) {
         HttpHeaders headers = new HttpHeaders();
         headers.set("Authorization", "Bearer " + jwtService.generateAccessToken(username));
-        headers.set("Refresh Token", jwtService.generateRefreshToken(username));
+        headers.set("X-Refresh-Token", jwtService.generateRefreshToken(username));
         return headers;
     }
 }


Daggerpov:
I think I'm parsing it fine on mobile

`			  let refreshToken = response.allHeaderFields["Refresh Token"] as? String else {`

Daggerpov:
> I think I'm parsing it fine on mobile
> 
> ` let refreshToken = response.allHeaderFields["Refresh Token"] as? String else {`

Whole method, for context:

```
private func handleAuthTokens(from response: HTTPURLResponse, for url: URL) throws {
		// Check if this is an auth endpoint
		let authEndpoints = [
			APIService.baseURL + "auth/sign-in",
			APIService.baseURL + "auth/make-user"
		]

		guard authEndpoints.contains(where: { url.absoluteString.contains($0) }),
			  let accessToken = response.allHeaderFields["Authorization"] as? String,
			  let refreshToken = response.allHeaderFields["Refresh Token"] as? String else {
			return
		}

		// Remove "Bearer " prefix from access token
		let cleanAccessToken = accessToken.replacingOccurrences(of: "Bearer ", with: "")

		// Store both tokens in keychain
		if let accessTokenData = cleanAccessToken.data(using: .utf8),
		   let refreshTokenData = refreshToken.data(using: .utf8) {
			if !KeychainService.shared.save(key: "accessToken", data: accessTokenData) {
				throw APIError.failedTokenSaving(tokenType: "accessToken")
			}
			if !KeychainService.shared.save(key: "refreshToken", data: refreshTokenData) {
				throw APIError.failedTokenSaving(tokenType: "refreshToken")
			}
		}
	}
```

Daggerpov (approved):
Ok, will test on mobile to see if that fixes current issue (will adjust mobile code to take `X-Refresh-Token`

PR #221, opened 2025-02-26
Fix event issue

ShaneMander:
The issue with fetching events was that `EventService::sortEventsByStartTime` was trying to sort (and thereby modify) an immutable list returned by `EventService::removeExpiredEvents`.

For reference, the following methods return immutable collections:
- `stream().collect(Collectors.toUnmodifiableList())`
- `stream().collect(Collectors.toList())`
- `stream().toList()`
- `List.of()`
- `Arrays.asList()`


Daggerpov (approved):
Hey, thanks for the reference. Same issue was happening earlier in another spot, so it’s great to now know the causes.

PR #223, opened 2025-02-26
Null Event Bugfix

evannawfal:
Adjusted getFeedEvents in EventService to filter out null events before calling removeExpiredEvents.


src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -444,6 +444,10 @@ public List<FullFeedEventDTO> getFeedEvents(UUID requestingUserId) {
                     convertEventsToFullFeedSelfOwnedEvents(getEventsByOwnerId(requestingUserId), requestingUserId);
             List<FullFeedEventDTO> eventsInvitedTo = getFullEventsInvitedTo(requestingUserId);
 
+            // Temp fix to remove null events
+            eventsCreated.removeIf(Objects::isNull);
```
Daggerpov, line 448: What do you think about one of these other solutions?
- Putting the null check in removeExpiredEvents()
- Null-checking the events lists returned itself is null, instead of individual event objects inside of it
    - I think this is the better check we should be doing, since those methods of getFullEventsInvitedTo() return null, not an events list which may contain nulls inside of it

Daggerpov (changes requested):
Food for thought 👅👅👅👅

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
evannawfal: Makes sense, I updated it to the better null handling + some more null checking in EventService

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -471,11 +475,19 @@ public List<FullFeedEventDTO> getFeedEvents(UUID requestingUserId) {
...
+        // If the entire list is null, return an empty list
+        if (events == null) {
+            return Collections.emptyList();
+        }
+
+        // Remove null events & filter out expired ones
         return events.stream()
-                .filter(event -> event.getEndTime() == null || !event.getEndTime().isBefore(now))
-                .collect(Collectors.toCollection(ArrayList::new));
+                .filter(event -> event != null && (event.getEndTime() == null || !event.getEndTime().isBefore(now)))
```
Daggerpov, line 486: Doesn’t this undo the fix @ShaneMander did in #221 ? I could be wrong, but please read his PR’s description & code to make sure

Daggerpov: I'm referring not to the `.filter()` line,. but the `.collect()` -> `.toList()` line btw. I think what you're doing with `event != null` is great, but we shouldn't touch the `.collect()` line as well

Daggerpov (approved):
Seems good. Let's watch out in the logs for a possible immutability issue with lists, as @ShaneMander pointed out

Issue #207, opened 2025-02-23
Add property to `FullFeedEventDTO`: `isSelfOwned`


Daggerpov:
On mobile, we need a way to distinguish self-owned events from ones a user has been invited to/is participating in, for our UI, as you can see below from our re-branded figma. So, this property would be the simplest way to accomplish this, so that on mobile we can check `if (event.isSelfOwned)` then adjust the colors accordingly. Since, this isn't just for the event color (which is already stored), but also for the surrounding elements. 

Alternatively, we could check on mobile if the color corresponds to the self-owned, but that'd be a bad practice of coupling logic, with having the color of an event dictate whether it's self-owned (could pose problems in the future, with custom event colors).

PR #229, opened 2025-02-28
added isSelfOwned field to FullFeedEventDTO

OnionFastener:
relatively simple fix to #207


OnionFastener:
Should i create another dummy event DTO where isSelfOwned is True?

Daggerpov (approved):
Looks great. To address your question, of course more tests are always nice to have

Issue #231, opened 2025-03-01
Remove Unnecessary DTO Properties


Daggerpov:
- [x] `FullFriendUserDTO` doesn't need:
    - [x]`friendTags`
    - [x] `friends`
    - [x] to extend from `FullUserDTO`, probably
- [x] `RecommendedFriendUserDTO` doesn't need:
    - [x] `friendTags`
    - [x] `friends`
    - [x] to extend from `FullUserDTO`, probably

- [x] `BaseUserDTO`, as discussed on Slack, per Shane
    - [x] FullFeedEventDTO
        - [x] creatorUser
        - [x] participantUsers
        - [x] invitedUsers
    - [x] FullFriendTagDTO
        - [x] friends
    
- [x] Make a FriendTagCreationDTO with only these properties:
    - [ ]`id: UUID, displayName: String, colorHexCode: String, ownerUserId: UUID`
    - [x] `isEveryone` should always be set to false when inserting into our DB 
    - I was already doing this on mobile by manually setting empty list and false for the two unneeded properties, but now we can just remove them if we use a new DTO at this endpoint: `/friendTags` (POST)

- [x] Replace `PotentialFriendUserDTO` and `AbstractFriendUserDTO` with `BaseUserDTO`

<img width="667" alt="Image" src="https://github.com/user-attachments/assets/3df91256-4c32-44b7-ae3e-e110edb34431" />

Daggerpov:
All done by @ShaneMander !!

PR #232, opened 2025-03-01
Remove Unnecessary DTO properties

ShaneMander:
For #231.

Introduced `AbstractFriendUserDTO` which both `RecommendedFriendUserDTO` and `FullFriendUserDTO` extend from, and removed unnecessary fields from each.


Daggerpov (approved):
I also thought we might want to add another class like this. Good work!

PR #236, opened 2025-03-01
Added PotentialFriendUserDTO

ShaneMander:
Like discussed in Slack, added a new `PotentialFriendUserDTO` and refactored `FetchFriendRequestDTO.senderUser` to be of type `PotentialFriendUserDTO`.

Also since the number of `FriendUserDTO`s are growing, I added a new sub-package of `DTOs.User` to encapsulate them: `DTOs.User.FriendUser`.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/DTOs/User/FriendUser/PotentialFriendUserDTO.java
+++ b/src/main/java/com/danielagapov/spawn/DTOs/User/FriendUser/PotentialFriendUserDTO.java
@@ -1,6 +1,5 @@
 package com.danielagapov.spawn.DTOs.User.FriendUser;
 
-import com.danielagapov.spawn.Models.User;
 import lombok.Getter;
 import lombok.Setter;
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/danielagapov/spawn/DTOs/User/FriendUser/PotentialFriendUserDTO.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.UUID;
+
+@Getter
+@Setter
+public class PotentialFriendUserDTO extends AbstractFriendUserDTO {
+    public PotentialFriendUserDTO(UUID id, String firstName, String lastName, String email, String username, String bio, String profilePicture) {
+        super(id, firstName, lastName, email, username, bio, profilePicture);
+    }
+
+    public static PotentialFriendUserDTO fromUserEntity(User user) {
```
Daggerpov, line 16: Would've preferred to have this in a mapper file; so after we merge, changing that would be nice.

Daggerpov (approved):
Good to merge, but one clean-up thing could be done later.

Daggerpov (changes requested):
Mapper class for method, instead of embedding and coupling in DTO

Daggerpov (approved):

PR #237, opened 2025-03-01
Optimize query

ShaneMander:
Refactored `FriendRequestService::getIncomingFetchFriendRequestsByUserId` (called with the request `GET /api/v1/friend-requests/incoming/{userId}`) so that we are now converting from `FriendRequest` entity directly to `FetchFriendRequestDTO` which ensures only 1 db query total. Previously we were converting `CreateFriendRequestDTO` to `FetchFriendRequestDTO` which meant 1 db query for each friend request a user has.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/DTOs/FriendRequest/FetchFriendRequestDTO.java
+++ b/src/main/java/com/danielagapov/spawn/DTOs/FriendRequest/FetchFriendRequestDTO.java
@@ -1,13 +1,10 @@
 package com.danielagapov.spawn.DTOs.FriendRequest;
 
 import com.danielagapov.spawn.DTOs.User.FriendUser.PotentialFriendUserDTO;
-import com.danielagapov.spawn.Models.FriendRequest;
-import com.danielagapov.spawn.Models.User;
 import lombok.Getter;
 import lombok.Setter;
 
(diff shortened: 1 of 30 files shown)


Daggerpov:
Changing base to see more accurate diff

src/main/java/com/danielagapov/spawn/DTOs/FriendRequest/FetchFriendRequestDTO.java
```diff
@@ -17,4 +20,13 @@ public FetchFriendRequestDTO(UUID id, PotentialFriendUserDTO senderUser) {
         super(id);
         this.senderUser = senderUser;
     }
-}
+
+    public static FetchFriendRequestDTO fromEntity(FriendRequest friendRequest) {
```
Daggerpov, line 24: Put in a mapper class

Daggerpov (approved):
Great change; it's good to go (same as other PR), but would prefer a mapper class to do dto<>entity conversions.

Daggerpov (changes requested):
Mapper class for method, instead of embedding and coupling in DTO

Daggerpov (dismissed):

Daggerpov (approved):

PR #239, opened 2025-03-02
Introduce BaseUserDTO

ShaneMander:
Addressed additional points in #231.


Daggerpov (changes requested):
I believe that with the addition of `BaseUserDTO`, we can now delete `PotentialFriendUserDTO` and `AbstractFriendUserDTO`, replacing them with `BaseUserDTO`s, as all 3 have the exact same properties.

Daggerpov:
Could you also replace the repo README image with an updated UML diagram of the User DTOs, through IntelliJ's diagrams feature?

src/main/java/com/danielagapov/spawn/DTOs/Event/FullFeedEventDTO.java
```diff
@@ -32,13 +32,12 @@ public FullFeedEventDTO(UUID id,
...
-                            UserDTO creatorUser,
-                            List<UserDTO> participantUsers,
-                            List<UserDTO> invitedUsers,
+                            BaseUserDTO creatorUser,
+                            List<BaseUserDTO> participantUsers,
+                            List<BaseUserDTO> invitedUsers,
                             List<FullEventChatMessageDTO> chatMessages,
                             String eventFriendTagColorHexCodeForRequestingUser,
-                            ParticipationStatus participationStatus,
-                            boolean isSelfOwned) {
```
Daggerpov, line 41: Why did you remove `isSelfOwned` from Owen's work in #229 ?

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -562,8 +558,7 @@ public FullFeedEventDTO getFullEventByEvent(EventDTO event, UUID requestingUserI
                     userService.getInvitedByEventId(event.getId()),
                     chatMessageService.getFullChatMessagesByEventId(event.getId()),
                     requestingUserId != null ? getFriendTagColorHexCodeForRequestingUser(event, requestingUserId) : null,
-                    requestingUserId != null ? getParticipationStatus(event.getId(), requestingUserId) : null,
-                    requestingUserId == event.getCreatorUserId()
+                    requestingUserId != null ? getParticipationStatus(event.getId(), requestingUserId) : null
```
Daggerpov, line 561: Revisit this if we add back `isSelfOwned`

img.png
Daggerpov, line 1: Let's put this and other images into an appropriate folder. They shouldn't have been in our root directory in the first place

src/test/java/com/danielagapov/spawn/ServiceTests/EventServiceTests.java
```diff
@@ -115,8 +115,7 @@ private FullFeedEventDTO dummyFullFeedEventDTO(UUID eventId, String title) {
                 List.of(),
                 List.of(),
                 null,
-                null,
-                false
```
Daggerpov, line 119: revisit if we add back `isSelfOwned`

Daggerpov (changes requested):

src/main/java/com/danielagapov/spawn/DTOs/Event/FullFeedEventDTO.java
ShaneMander: It was unused

Daggerpov: Please revert it. I might've made it useful in a future commit that you haven't pulled into this branch.

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
```diff
@@ -157,12 +154,7 @@ private FullUserDTO getFullUserDTO(UUID externalUserId) {
 
     private void createAndSaveMapping(String externalUserId, AbstractUserDTO userDTO, OAuthProvider provider) {
         try {
-            User user;
-            if (userDTO instanceof FullUserDTO) {
-                user = UserMapper.convertFullUserToUserEntity((FullUserDTO) userDTO);
-            } else {
-                user = UserMapper.toEntity((UserDTO) userDTO);
-            }
+            User user = UserMapper.toEntity((BaseUserDTO) userDTO);
```
Daggerpov, line 157: Are you sure this method only takes in `BaseUserDTO`s? Seems like its only usage is being given a `UserDTO`

image

image

Daggerpov (changes requested):
Additional possible bug

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
ShaneMander: Right, a long time ago this method sometimes used to take in a FullUserDTO instead. I don't see how it may cause a bug, because the `BaseUserDTO toEntity` mapper works the same for either a FullUserDTO or UserDTO. In any case, I've changed the signature of `createAndSaveMapping` to take in a `UserDTO` so it is more accurate to its current usage.

Daggerpov (approved):
Solid! Thanks for the additional code & image clean-up too!

src/main/java/com/danielagapov/spawn/Services/OAuth/OAuthService.java
Daggerpov: Ah, right, because they both inherit. Good point! Thanks for changing it anyway!

Daggerpov (approved):
👏👏👏

PR #240, opened 2025-03-02
Friendtag Duplicate Bugfix

evannawfal:
Added a check in saveUserToFriendTag to verify if the user already exists before inserting, preventing the error we were seeing.


Daggerpov (approved):
Makes sense!

PR #247, opened 2025-03-06
Fix sign-in bug and introduce ErrorResponse

ShaneMander:
Updated Oauth sign-in so that an error message is sent back when a user tries to log into their account with same email but different provider.

This should also resolve a bug with the sign-in endpoint.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/User/UserService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/User/UserService.java
@@ -298,6 +298,22 @@ public class UserService implements IUserService {
     }
 
     @Override
+    public FullUserDTO getFullUserByUserEntity(User user) {
+        return getFullUserByUser(getUserDTOByEntity(user), Set.of());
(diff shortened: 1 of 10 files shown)


src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -297,6 +297,22 @@ public FullUserDTO getFullUserByEmail(String email) {
         }
     }
 
+    @Override
+    public FullUserDTO getFullUserByUserEntity(User user) {
+        return getFullUserByUser(getUserDTOByEntity(user), Set.of());
+    }
+
+    @Override
+    public UserDTO getUserDTOByEntity(User user) {
```
Daggerpov, line 306: Ah, this is the optimization you were talking about with already having some properties computed. Makes sense!

Daggerpov (approved):
Everything makes a lot of sense! Great work!

PR #248, opened 2025-03-06
Log levels

ShaneMander:
Thoughts on adding log levels? Should hopefully make our logs on railway easier to look through.

IntelliJ:
image

Railway: 
image


src/main/java/com/danielagapov/spawn/Exceptions/Logger/Logger.java
```diff
@@ -3,7 +3,21 @@
...
+
+    public void warn(String message) {
+        logger.warn(message);
+    }
+
+    public void error(String message) {
+        logger.error(message);
+    }
+
     public void log(String message) {
```
Daggerpov, line 21: Is it possible to have our current `logger.log` calls go to the new `logger` (from `org.slf4j.Logger`)'s `.warn()` and `.error()` in this PR?

Daggerpov (changes requested):
Great idea; I'd love to implement it. I think since staging is waiting to be merged into main, we may as well finish the job in this PR of making the current `.log(e.getMessage())` calls go to the `.error()` method, and other logs go to `.info()`

src/main/java/com/danielagapov/spawn/Exceptions/Logger/Logger.java
ShaneMander: Yeah I can do that

ShaneMander:
@Daggerpov `logger.log()` has now been replaced by `.info()`, `.warn()`, and `.error()`. Here is the log-level usage I tried to adhere to:

- `info()` - used for logging general statements like "Getting full feed events for requesting user id..."
- `.warn()` - used for logging statements that accompany exceptions that are handled. For example, if an IncorrectProviderException is thrown then I would use a warn log statement, if I wanted to log at all.
- `.error()` - used for logging statements that accompany exceptions that are not caught, i.e. any exception that would lead to a status code 500. 

I highly recommend not reviewing this PR, but rather merging and then taking note of any incorrect log statements (e.g. `.info()` where it should be `.error()`) and correcting them as you see them, or reaching out to me.

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -19,10 +20,12 @@
 public class UserController {
     private final IUserService userService;
     private final IS3Service s3Service;
+    private final ILogger logger;
 
-    public UserController(IUserService userService, IS3Service s3Service) {
+    public UserController(IUserService userService, IS3Service s3Service, ILogger logger) {
```
Daggerpov, line 25: Seems to be no reason to add this dependency

src/main/java/com/danielagapov/spawn/Exceptions/Logger/Logger.java
```diff
@@ -3,17 +3,18 @@
...
+        logger.info(message);
+    }
+
+    public void warn(String message) {
+        logger.warn(message);
+    }
 
-        // Prints the details in a readable format ([FileName:LineNumber] MethodName - Message)
-        System.out.printf("[%s:%d] %s - %s%n", fileName, lineNumber, methodName, message);
+    public void error(String message) {
```
Daggerpov, line 17: I guess there's no longer a reason to have our own `Logger`, but it could be useful in the future if we want to extend/modify the imported logger you have. For example, with the custom print statement I had earlier if we see fit. For now, this is fine.

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
```diff
@@ -134,10 +148,10 @@ public ChatMessageDTO saveChatMessage(ChatMessageDTO chatMessageDTO) {
```
Daggerpov, line 148: is `saveChatMessage()` being used anymore? If not, let's remove it.

src/main/java/com/danielagapov/spawn/Exceptions/Logger/Logger.java
Daggerpov: And, now I'm seeing that we have your `.xml` file to control the console logging, so this `Logger` class of our own is probably useless (but, again, not a priority to remove for now, since it could bring use).

Daggerpov (approved):
Left a few clean-up comments that can be addressed later. Let's not hold up this PR, since it contains #251

PR #250, opened 2025-03-07
OAuth BaseUserDTO refactor

ShaneMander:
As discussed on Slack, refactored the endpoints `AuthController::makeUser` and `AuthController::signIn` to return `BaseUserDTO`s instead of `FullUserDTO`s. Also updated the tests accordingly.


Daggerpov (approved):
Solid!

PR #251, opened 2025-03-07
CreateChatMessageDTO

ShaneMander:
As requested on Slack, introduced a new `CreateChatMessageDTO` which is used in the request body for creating a new chat message (`POST /chatMessages`)


Daggerpov (approved):
Solid. Wish it hadn't been branched from your other PR so we could merge it right away.

PR #253, opened 2025-03-08
Everyone Tag Deletion Fix

evannawfal:
Blocked deletion for the Everyone tag.


Daggerpov (approved):
Nice block. I've also added front-end code to make it impossible:

image

Daggerpov (approved):

Issue #185, opened 2025-02-18
Apple Push Notifications


Daggerpov:
https://victorleungtw.medium.com/apple-push-notification-with-java-spring-framework-b6ddada216cf

ShaneMander:
Hey, let me know if you need any backend-related help with setting up the notifications. I was doing a bit of research the other day on notifications and it seems kinda interesting/fun.

PR #255, opened 2025-03-08
Push Notifications

Daggerpov:
First pass at push notifications. Implemented the notification behavior as described in #185.

This is mostly based off of Claude's work, with some tweaks and additions.


ShaneMander:
Cool, I'll look through

src/main/resources/application.properties
```diff
@@ -11,4 +11,11 @@ spring.mail.username=spawnappmarketing@gmail.com
 spring.mail.password=${EMAIL_PASS}
 spring.mail.properties.mail.smtp.auth=true
 spring.mail.properties.mail.smtp.starttls.enable=true
-spring.mail.properties.mail.smtp.ssl.trust=smtp.gmail.com
\ No newline at end of file
+spring.mail.properties.mail.smtp.ssl.trust=smtp.gmail.com
+# Push Notification Configuration
+# APNS Configuration for iOS
+apns.certificate.path=${APNS_CERTIFICATE}
```
Daggerpov, line 17: Because we have these env vars here, let's push to `staging` to see if missing them will crash the railway deployment

ShaneMander:
I have temporarily disabled some of the failing tests. They are currently failing because DB calls from the new notification methods in each service class are not returning anything. I am disabling because I believe they will be resolved after the notification refactor we had discussed, so we can properly mock these calls.

ShaneMander (approved):
Feel free to merge when you are ready from the mobile side

PR #257, opened 2025-03-09
Endpoint Mobile Usage Documentation

Daggerpov:
- Removed all mock endpoints
- Documented unused endpoints with `@Deprecated`
    - If those endpoints will soon be used by mobile, as documented in [this mobile issue](https://github.com/Daggerpov/Spawn-App-iOS-SwiftUI/issues/142), then they mention that

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/ChatMessageController.java
@@ -27,10 +27,7 @@ public class ChatMessageController {
     }
 
 
-    // Despite this not currently being used, let's not delete it.
-    // Recently, I noticed that our `EventController::getFullEventById()` endpoint
-    // wasn't being used -> so I removed it, but had to add it back for a new feature.
-    // So, let's consider removing it in the future, but keep for now. 
+    // TL;DR: Don't remove this endpoint; it may become useful. 
     @Deprecated(since = "Not being used on mobile currently.")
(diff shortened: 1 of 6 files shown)


ShaneMander (changes requested):
Thanks, it's helpful to know what is used by mobile!

I don't think we need the block comment on every single deprecated endpoint, since it clutters the controller classes a bit. Can we either leave the comment on one of the deprecated endpoints or once at the top of each class?

Daggerpov:
> Thanks, it's helpful to know what is used by mobile!
> 
> I don't think we need the block comment on every single deprecated endpoint, since it clutters the controller classes a bit. Can we either leave the comment on one of the deprecated endpoints or once at the top of each class?

Good point about it being blocky; I made it a "TL;DR" summary. Plus, I can always review the PRs to stop someone from deleting it. I just want it to be obvious for that method it shouldn't be deleted.

ShaneMander:
> > Thanks, it's helpful to know what is used by mobile!
> > I don't think we need the block comment on every single deprecated endpoint, since it clutters the controller classes a bit. Can we either leave the comment on one of the deprecated endpoints or once at the top of each class?
> 
> Good point about it being blocky; I made it a "TL;DR" summary. Plus, I can always review the PRs to stop someone from deleting it. I just want it to be obvious for that method it shouldn't be deleted.

Yup I think the TL;DR should suffice. Thanks

ShaneMander (approved):
👍

PR #215, opened 2025-02-24
Feedback submission/retrieval

RaulUrbina:
I implemented the endpoints related to feedback submission, following the current codebase structure. I hope it's okay!

Issue #196 

### 1. Submit Feedback

Endpoint:
`POST /api/v1/feedback`

Request Body:
```
{
  "type": "BUG_REPORT", // Enum: BUG_REPORT, FEATURE_REQUEST, GENERAL_FEEDBACK
  "fromUserId": "123e4567-e89b-12d3-a456-426614174000", // Optional: UUID of the user submitting the feedback
  "fromUserEmail": "user@example.com", // Optional: Email of the user submitting the feedback (Either fromUserId or fromUserEmail must be submitted)
  "message": "The application crashes when clicking the submit button." // Required: Feedback message
}
``` 

### 2. Retrieve All Feedback Submissions

Endpoint:
`GET /api/v1/feedback`

Example Response:
```
[
  {
    "id": "123e4567-e89b-12d3-a456-426614174000",
    "type": "BUG_REPORT",
    "fromUserId": "123e4567-e89b-12d3-a456-426614174000",
    "submittedAt": "2023-10-25T14:30:00Z",
    "message": "The application crashes when clicking the submit button."
  },
  {
    "id": "223e4567-e89b-12d3-a456-426614174000",
    "type": "FEATURE_REQUEST",
    "fromUserId": "223e4567-e89b-12d3-a456-426614174000",
    "submittedAt": "2023-10-25T15:00:00Z",
    "message": "Add dark mode support."
  }
]
```

### Entity Created
image

Let me know if you need any adjustment/improvements :D

PR #260, opened 2025-03-10
Clean Feedback Submission/Retrieval Fix

evannawfal:
Pushed to different branch because of compiling errors.

Resolves #215


src/main/java/com/danielagapov/spawn/Models/FeedbackSubmission.java
```diff
@@ -0,0 +1,43 @@
...
+@Setter
+public class FeedbackSubmission implements Serializable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private UUID id;
+
+    @Enumerated(EnumType.STRING)
+    private FeedbackType type;
+
+    private UUID fromUserId;
```
ShaneMander, line 27: Can we change this to a foreign key relationship like in other Model classes? That way we can also add `@OnDelete(action = OnDeleteAction.SET_NULL)` to the user property which will automatically set the field to null if a user's account gets deleted.

src/main/java/com/danielagapov/spawn/Services/FeedbackSubmission/FeedbackSubmissionService.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            FeedbackSubmission feedback = FeedbackSubmissionMapper.toEntity(dto);
+            feedback.setFromUserId(userId);
+
+            if (feedback.getFromUserEmail() == null) {
+                feedback.setFromUserEmail(dto.getFromUserEmail());
+            }
+
+            return repository.save(feedback);
```
ShaneMander, line 63: Feel free to correct me if I'm wrong but I think we can simplify this to just checking the existence of a user, i.e. `userRepository.existsById`, then saving the feedback if a user exists, otherwise throwing an error if a user doesn't exist. Emails can sometimes be null, but user ids should never be null.

src/main/java/com/danielagapov/spawn/Models/FeedbackSubmission.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Column
+    private String fromUserEmail;
+    private OffsetDateTime submittedAt;
+
```
ShaneMander, line 32: I think it could be useful to also add some sort of boolean property like `isResolved` which we can use to internally monitor and track which feedback reports have been addressed.

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
```diff
@@ -0,0 +1,56 @@
...
+    public ResponseEntity<FeedbackSubmission> submitFeedback(@RequestBody FeedbackSubmissionDTO dto) {
+        try {
+            return new ResponseEntity<>(service.submitFeedback(dto), HttpStatus.CREATED);
+        } catch (BaseSaveException e) {
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        } catch (Exception e) {
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
```
ShaneMander, line 40: Related to my comment about `isResolved`: we would also need an endpoint like `PUT feedback/resolve` (very similar to `ReportController::resolveReport`) so we can mark feedback reports as resolved.

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
```diff
@@ -0,0 +1,56 @@
...
+@RestController
+@RequestMapping("api/v1/feedback")
+public class FeedbackSubmissionController {
+    private final IFeedbackSubmissionService service;
+
+    public FeedbackSubmissionController(IFeedbackSubmissionService service) {
+        this.service = service;
+    }
+
```
ShaneMander, line 23: Might need a delete endpoint eventually as well

ShaneMander (changes requested):
Left some comments about modifying `FeedbackSubmission` entity to include a FK relationship with `User` and to add an `isResolved` property

Daggerpov:
Once this merges, feel free to close Raul's PR #215

src/main/java/com/danielagapov/spawn/Services/FeedbackSubmission/FeedbackSubmissionService.java
Daggerpov: Yea, I guess Raul wasn't familiar with our codebase to know that we rely more on user ids than emails

src/main/java/com/danielagapov/spawn/Models/FeedbackSubmission.java
Daggerpov: Plus, we should have a `resolutionComment` field to comment on the reason, like by saying "duplicate feature request to something already documented in Notion" or "resolved since it's a bug already addressed since this was submitted"

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
Daggerpov: Can make it once we need it. I'll resolve this for now

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -107,23 +109,6 @@ public ResponseEntity<UserDTO> createUser(@RequestParam("user") UserDTO newUser,
         }
     }
 
-    // TL;DR: Don't remove this endpoint; it may become useful.
-    @Deprecated(since = "Not being used on mobile currently. " +
```
ShaneMander, line 111: Did @Daggerpov ask to remove this endpoint? I thought we wanted to leave this in?

ShaneMander: Oh is this from another branch or smt?

src/main/java/com/danielagapov/spawn/DTOs/FeedbackSubmissionDTO.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.UUID;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class FeedbackSubmissionDTO {
```
ShaneMander, line 13: I think this is missing `implements Serializable`

src/main/java/com/danielagapov/spawn/DTOs/User/UserUpdateDTO.java
```diff
@@ -0,0 +1,15 @@
+package com.danielagapov.spawn.DTOs.User;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class UserUpdateDTO {
```
ShaneMander, line 10: Should implement Serializable

ShaneMander: I think we are also missing an id field, e.g. `private UUID id`

src/main/java/com/danielagapov/spawn/DTOs/FeedbackSubmissionDTO.java
ShaneMander: Also missing an id field

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
```diff
@@ -0,0 +1,97 @@
...
+    // Full path: /api/v1/feedback/resolve/{id}
+/**
+     * Endpoint for resolving a feedback submission.
+     * @param id The unique ID of the feedback submission to resolve.
+     * @param resolutionComment An optional comment to add when resolving the feedback.
+     * @return A success response if the feedback was resolved, otherwise an error response.
+     */
+    @PutMapping("/resolve/{id}")
+    public ResponseEntity<FeedbackSubmissionDTO> resolveFeedback(
+            @PathVariable UUID id,
```
ShaneMander, line 51: This id PathVariable can be removed when we add the id field to `FeedbackSubmissionDTO`. The id field is needed in the DTO because as far as I can tell, the client here would never actually know what the id of the feedback entity is, if it isn't present in the DTO that we provide

src/main/java/com/danielagapov/spawn/Services/FeedbackSubmission/FeedbackSubmissionService.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    @Override
+    public FeedbackSubmissionDTO deleteFeedback(UUID id) {
+        FeedbackSubmission feedback = repository.findById(id)
+                .orElseThrow(() -> new BaseNotFoundException(EntityType.FeedbackSubmission, id));
+
+        repository.delete(feedback);
+        return FeedbackSubmissionMapper.toDTO(feedback);
```
ShaneMander, line 100: I think this method can be just simplified to `repository.deleteById` which is slightly better for performance reasons. We also don't need to send back a `FeedbackSubmissionDTO` here because it's being deleted. I think it would be better to change the return type to either `void` or `boolean` (true in successful deletes, false when there are exceptions).

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    /**
+     * Endpoint for deleting a feedback submission.
+     * @param id The unique ID of the feedback submission to delete.
+     * @return A success response if the feedback was deleted, otherwise an error response.
+     */
+    @DeleteMapping("/delete/{id}")
+    public ResponseEntity<FeedbackSubmissionDTO> deleteFeedback(@PathVariable UUID id) {
```
ShaneMander, line 87: I don't think we should send back the deleted `FeedbackSubmissionDTO` because its being deleted. The return type could just probably just be `Void` or `?` if you want to sometimes send back error messages. Up to you

src/main/java/com/danielagapov/spawn/Services/FriendRequest/FriendRequestService.java
```diff
@@ -62,7 +62,10 @@ public CreateFriendRequestDTO saveFriendRequest(CreateFriendRequestDTO friendReq
     @Override
     public List<FetchFriendRequestDTO> getIncomingFetchFriendRequestsByUserId(UUID id) {
         List<FriendRequest> friendRequests = getIncomingFriendRequestsByUserId(id);
-        return FetchFriendRequestMapper.toDTOList(friendRequests);
```
ShaneMander, line 65: Why did we delete this call?

ShaneMander (changes requested):
Nice stuff. I just left a few small comments on missing an `id field and implements Serializable` in the DTOs and some return types. Should be good to go after that.

Also, I'm not sure if I've commented on something that was someone else's work from another branch. If I did then you can just ignore it.

src/main/java/com/danielagapov/spawn/DTOs/FeedbackSubmissionDTO.java
Daggerpov: Done and done

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
Daggerpov: Since id has just been added to `FeedbackSubmissionDTO`, we can keep this endpoint as-is, since the client (us on an admin website or page within the app) would load the feedback submission, and thus know the id.

src/main/java/com/danielagapov/spawn/Services/FeedbackSubmission/FeedbackSubmissionService.java
Daggerpov: Addressed by making it return `void`, and the controller deals with the errors as usual.

src/main/java/com/danielagapov/spawn/Controllers/FeedbackSubmissionController.java
Daggerpov: Yea, didn't make sense. I'm sending back 204 no content (standard for deletion) or a 500 if error.

Daggerpov (approved):

PR #262, opened 2025-03-10
Cherry-picked mutual friend count addition

Daggerpov:
Created this separate PR so we don't have to merge #260 before this, since this is more important--likely fixes the bug in #testing-qa
- Replaced `.toDTOList()` with a direct `.map()` call from service class, to avoid having to inject a dependency (`UserService`) into the mapper class


ShaneMander (approved):
Looks good

Issue #224, opened 2025-02-27
(Huge Overhaul) Utilize Java Optionals instead of != null checking


Daggerpov:
Basically, more of what I did in here 2 Bug Fixes + Optionals for DB Queries #243 

This article helped a lot: https://www.baeldung.com/java-optional

Much safer option, to avoid `NullPointerException`s. This video details it well:

https://youtu.be/vKVzRbsMnTQ?si=7y3ZjI91EKG20ORH

Daggerpov:
Doing in #264

PR #264, opened 2025-03-10
More `Optional<>` surrounding of DB Queries + Clean-Up

Daggerpov:
#224 

This will safe-guard against more null pointer exceptions/invocations. 

Applies to all applicable repo methods, except for `List<>` returns, since @ShaneMander pointed out that it's redundant, since those will never return null values--as they will return empty lists by default through JPA. 

Also removed redundant imports, variables, and `implements Serializable` on DTOs (that inherit them from their parent DTO classes).

### DIFF
--- a/src/main/java/com/danielagapov/spawn/DTOs/Event/EventCreationDTO.java
+++ b/src/main/java/com/danielagapov/spawn/DTOs/Event/EventCreationDTO.java
@@ -3,14 +3,13 @@ package com.danielagapov.spawn.DTOs.Event;
 import lombok.Getter;
 import lombok.Setter;
 
-import java.io.Serializable;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.UUID;
 
 @Getter
 @Setter
-public class EventCreationDTO extends AbstractEventDTO implements Serializable {
+public class EventCreationDTO extends AbstractEventDTO {
     LocationDTO location;
     UUID creatorUserId;
     List<UUID> invitedFriendTagIds;
--- a/src/main/java/com/danielagapov/spawn/Repositories/IChatMessageRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IChatMessageRepository.java
@@ -5,9 +5,10 @@ import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 @Repository
 public interface IChatMessageRepository extends JpaRepository<ChatMessage, UUID> {
-    List<ChatMessage> getChatMessagesByEventId(UUID eventId);
+    Optional<List<ChatMessage>> getChatMessagesByEventId(UUID eventId);
 }
--- a/src/main/java/com/danielagapov/spawn/Repositories/IChatMessageRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IChatMessageRepository.java
@@ -5,10 +5,9 @@ import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.UUID;
 
 @Repository
 public interface IChatMessageRepository extends JpaRepository<ChatMessage, UUID> {
-    Optional<List<ChatMessage>> getChatMessagesByEventId(UUID eventId);
+    List<ChatMessage> getChatMessagesByEventId(UUID eventId);
 }
--- a/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
@@ -5,9 +5,10 @@ import com.danielagapov.spawn.Models.EventUser;
 import org.springframework.data.jpa.repository.JpaRepository;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 public interface IEventUserRepository extends JpaRepository<EventUser, EventUsersId> {
-    List<EventUser> findByEvent_Id(UUID eventId);
-    List<EventUser> findByUser_Id(UUID userId);
+    Optional<List<EventUser>> findByEvent_Id(UUID eventId);
+    Optional<List<EventUser>> findByUser_Id(UUID userId);
 }
--- a/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
@@ -10,15 +10,15 @@ import java.util.Optional;
 import java.util.UUID;
 
 public interface IEventUserRepository extends JpaRepository<EventUser, EventUsersId> {
-    Optional<List<EventUser>> findByEvent_Id(UUID eventId);
-    Optional<List<EventUser>> findByUser_Id(UUID userId);
+    List<EventUser> findByEvent_Id(UUID eventId);
+    List<EventUser> findByUser_Id(UUID userId);
     
     // Find event users by event ID and participation status
-    Optional<List<EventUser>> findByEvent_IdAndStatus(UUID eventId, ParticipationStatus status);
+    List<EventUser> findByEvent_IdAndStatus(UUID eventId, ParticipationStatus status);
     
     // Find event users by user ID and participation status
-    Optional<List<EventUser>> findByUser_IdAndStatus(UUID userId, ParticipationStatus status);
+    List<EventUser> findByUser_IdAndStatus(UUID userId, ParticipationStatus status);
     
     // Find event users by event ID and user ID
-    Optional<List<EventUser>> findByEvent_IdAndUser_Id(UUID eventId, UUID userId);
+    List<EventUser> findByEvent_IdAndUser_Id(UUID eventId, UUID userId);
 }
--- a/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
@@ -6,7 +6,6 @@ import com.danielagapov.spawn.Models.EventUser;
 import org.springframework.data.jpa.repository.JpaRepository;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.UUID;
 
 public interface IEventUserRepository extends JpaRepository<EventUser, EventUsersId> {
--- a/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
@@ -1,5 +1,6 @@
 package com.danielagapov.spawn.Repositories;
 
+import com.danielagapov.spawn.Enums.ParticipationStatus;
 import com.danielagapov.spawn.Models.CompositeKeys.EventUsersId;
 import com.danielagapov.spawn.Models.EventUser;
 import org.springframework.data.jpa.repository.JpaRepository;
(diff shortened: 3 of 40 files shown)


src/main/java/com/danielagapov/spawn/Repositories/IEventUserRepository.java
```diff
@@ -1,13 +1,24 @@
...
-    List<EventUser> findByEvent_Id(UUID eventId);
-    List<EventUser> findByUser_Id(UUID userId);
+    Optional<List<EventUser>> findByEvent_Id(UUID eventId);
+    Optional<List<EventUser>> findByUser_Id(UUID userId);
+    
+    // Find event users by event ID and participation status
+    Optional<List<EventUser>> findByEvent_IdAndStatus(UUID eventId, ParticipationStatus status);
+    
+    // Find event users by user ID and participation status
+    Optional<List<EventUser>> findByUser_IdAndStatus(UUID userId, ParticipationStatus status);
```
ShaneMander, line 20: Great idea to add `findByUser_IdAndStatus`

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -75,6 +77,15 @@ public interface IUserService {
 
     List<UUID> getInvitedUserIdsByEventId(UUID eventId);
 
+    // Event User methods:
```
ShaneMander, line 80: We should probably consider placing these methods in an `EventUserService` class

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -516,7 +514,8 @@ public List<RecommendedFriendUserDTO> getRecommendedFriendsForUserId(UUID userId
     @Override
     public List<BaseUserDTO> getParticipantsByEventId(UUID eventId) {
         try {
-            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId);
+            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId)
```
ShaneMander, line 517: Can use `EventUserRepository::findByEvent_IdAndStatus` here

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -528,11 +527,11 @@ public List<BaseUserDTO> getParticipantsByEventId(UUID eventId) {
         }
     }
 
-
     @Override
     public List<BaseUserDTO> getInvitedByEventId(UUID eventId) {
         try {
-            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId);
+            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId)
```
ShaneMander, line 533: Can also use `EventUserRepository::findByEvent_IdAndStatus` here as well

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -544,11 +543,11 @@ public List<BaseUserDTO> getInvitedByEventId(UUID eventId) {
         }
     }
 
-
     @Override
     public List<UUID> getParticipantUserIdsByEventId(UUID eventId) {
```
ShaneMander, line 547: Tripped out again -- I also just added this method to the notification branch

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -544,11 +543,11 @@ public List<BaseUserDTO> getInvitedByEventId(UUID eventId) {
...
         try {
-            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId);
+            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId)
```
ShaneMander, line 549: `EventUserRepository::findByEvent_IdAndStatus`

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
```diff
@@ -560,11 +559,11 @@ public List<UUID> getParticipantUserIdsByEventId(UUID eventId) {
         }
     }
 
-
     @Override
     public List<UUID> getInvitedUserIdsByEventId(UUID eventId) {
         try {
-            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId);
+            List<EventUser> eventUsers = eventUserRepository.findByEvent_Id(eventId)
```
ShaneMander, line 565: `EventUserRepository::findByEvent_IdAndStatus`

ShaneMander (changes requested):
Nice, I added some comments about places you missed where you could've used the newly added `EventUserRepository::findByEvent_IdAndStatus`. 

Additionally, according to [this](https://docs.spring.io/spring-data/jpa/reference/repositories/null-handling.html), JPA methods that return collections are guaranteed to never be null (see second paragraph)... so it looks like we don't need Optionals there. Not sure what you want to do about that, considering you just modified all of the methods to return Optionals but maybe in the future once our backend implementation feels more stable we can consider removing those Optionals so that calling JPA methods that return a list aren't unnecessarily verbose (`.orElse()` at every call).

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -75,6 +77,15 @@ public interface IUserService {
...
+    List<EventUser> getEventUsersByEventId(UUID eventId);
+
+    List<EventUser> getEventUsersByUserId(UUID userId);
+
+    List<EventUser> getEventUsersByUserIdAndStatus(UUID userId, ParticipationStatus status);
+
+    List<EventUser> getEventUsersByEventIdAndStatus(UUID eventId, ParticipationStatus status);
+
```
ShaneMander, line 88: I could've missed it but where are these used? It seems that `EventService` already has its own instance of `EventUserRepository` which it already uses to do what these methods do.

src/main/java/com/danielagapov/spawn/Repositories/IChatMessageRepository.java
```diff
@@ -5,9 +5,10 @@
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 @Repository
 public interface IChatMessageRepository extends JpaRepository<ChatMessage, UUID> {
-    List<ChatMessage> getChatMessagesByEventId(UUID eventId);
+    Optional<List<ChatMessage>> getChatMessagesByEventId(UUID eventId);
```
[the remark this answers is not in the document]
Daggerpov: Addressed here: https://github.com/Daggerpov/Spawn-App-Back-End/pull/264/commits/483a2ed036a6feb48f2db361aaead9629d92137e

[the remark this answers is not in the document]
Daggerpov: Will address for all other `Optional<List<` instances in JPA-extending repos (all of our repos) in future commit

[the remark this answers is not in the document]
Daggerpov: Done in this range of commits: https://github.com/Daggerpov/Spawn-App-Back-End/pull/264/files/483a2ed036a6feb48f2db361aaead9629d92137e..3673e3342d6533c10c17a1901f540b6d980c932f#diff-338426c910019182f91bc1bf3a58bd33d9abc8243f35238c44dc641f8d0a584e

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Done

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Done

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Done

src/main/java/com/danielagapov/spawn/Services/User/UserService.java
Daggerpov: Done

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -304,24 +295,11 @@ public boolean deleteEventById(UUID id) {
...
-            }
-
-            return eventUsers.stream()
-                    .filter(eventUser -> eventUser.getStatus().equals(ParticipationStatus.participating))
-                    .map(eventUser -> userService.getUserById(eventUser.getUser().getId()))
-                    .toList();
-        } catch (DataAccessException e) {
-            logger.error(e.getMessage());
-            throw new BaseNotFoundException(EntityType.Event, eventId);
-        } catch (Exception e) {
```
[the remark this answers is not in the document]
Daggerpov: Added back in https://github.com/Daggerpov/Spawn-App-Back-End/commit/e702dc9e517b076da51cced28eafae9a07f046e1

src/main/java/com/danielagapov/spawn/DTOs/Event/EventCreationDTO.java
```diff
@@ -3,14 +3,13 @@
...
 import lombok.Setter;
 
-import java.io.Serializable;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.UUID;
 
 @Getter
 @Setter
-public class EventCreationDTO extends AbstractEventDTO implements Serializable {
```
Daggerpov, line 13: Removed these `implements Serializable` calls since these dtos inherit them from their parent classes @ShaneMander

ShaneMander: Good catch

src/main/java/com/danielagapov/spawn/Repositories/IUserFriendTagRepository.java
```diff
@@ -19,5 +20,5 @@ public interface IUserFriendTagRepository extends JpaRepository<UserFriendTag, U
     @Query("DELETE FROM UserFriendTag uft WHERE uft.friendTag.id = :tagId AND uft.friend.id = :friendId")
     void deleteByFriendTagIdAndUserId(@Param("tagId") UUID tagId, @Param("friendId") UUID friendId);
 
-    boolean existsByFriendTagIdAndFriendId(UUID id, UUID userId);
+    Optional<Boolean> existsByFriendTagIdAndFriendId(UUID id, UUID userId);
```
ShaneMander, line 23: I think Java primitives cannot be null. For example, if you try to assign null to a variable of type `boolean`, it'll complain right away. It isn't as explicit as with Collection return types but in [this source](https://docs.spring.io/spring-data/jpa/reference/repositories/query-return-types-reference.html), compare the description for a primitive return type vs. an entity (`T`) return type (says an entity can be null but does not say anything about primitives returning null. ChatGPT also says that `boolean` return types will never be null but rather default to `false`, if you want to trust that.

ShaneMander (changes requested):
Nice cleanup with the DTOs. I left one comment about JPA methods and primitives, in particular, `boolean` return types, which I think also will not be null. Sorry I didn't mention it before along with Collection return types, it didn't occur to me until seeing it again now.

src/main/java/com/danielagapov/spawn/Repositories/IUserFriendTagRepository.java
Daggerpov: Makes sense; glad you've been looking into all of this documentation lately. I've addressed it here: https://github.com/Daggerpov/Spawn-App-Back-End/pull/264/commits/6a36953da4002c6c60b0527e2ececb5b3c82efcb

PR #266, opened 2025-03-11
Events by Friend Tag bug fix

ShaneMander:
Per Slack: "it seems the bug is due to the way events are fetched through `EventController::getEventsByFriendTag()`. Currently, `getEventsByFriendTag()` is getting events that a user's friends has created, when it should be events that a user has been invited to."

With this PR, we are now getting events that a user has been invited by all their friends who belong to the specified friend tag. This logic is done at the database level with a single query (`IEventRepository::getEventsInvitedToWithFriendTagId`).

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Repositories/IEventRepository.java
+++ b/src/main/java/com/danielagapov/spawn/Repositories/IEventRepository.java
@@ -2,6 +2,8 @@ package com.danielagapov.spawn.Repositories;
 
 import com.danielagapov.spawn.Models.Event;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/com/danielagapov/spawn/Repositories/IEventRepository.java
```diff
@@ -16,4 +18,12 @@ public interface IEventRepository extends JpaRepository<Event, UUID> {
 
     // finds events that have been created by users, whose ids are in the `creatorIds` list
     List<Event> findByCreatorIdIn(List<UUID> creatorIds);
+
+    @Query("SELECT e FROM Event e " +
```
Daggerpov, line 22: Let's try to avoid raw sql statements as much as possible, since they're not syntax checked here; it's more prone to bugs since not only the syntax isn't checked, but the action validity of the joins and such. Also, if we change the names of things, then this raw sql breaks in the future. 

Bad practice, but it's fine for now to get the job done

Daggerpov: Also, it just makes this logic harder to maintain, than through code, since we can't re-use it as easily for both the filtered events (by friend tag) and feed events endpoints

Daggerpov: Ideally, we'd have a method to get all events a user is invited to, then that's good enough for the feed events endpoint, and the filter by friend tag endpoint would call then method then filter

Daggerpov (approved):
Fine for now, but left refactoring suggestion

PR #268, opened 2025-03-12
Friend queries

ShaneMander:
Friend fetching has been slow on mobile recently. I've refactored some of that fetching to hopefully help improve performance. 

To do this I added a JPA method to the `FriendTag` repository to fetch all friends in a single query. Also, we are no longer needlessly converting users to `FullUserDTO`s (which likely got left behind when doing recent DTO refactors) in `UserService::getFullFriendUsersByUserId`.


Daggerpov (dismissed):
Makes sense that we don't need to retrieve full users in an intermediary step. Nice optimization!

PR #279, opened 2025-03-21
Refactoring Notifications -- Strategy & Observer Design Patterns

Daggerpov:
Utilizing the [strategy design pattern](https://www.baeldung.com/java-strategy-pattern) to define notification 'strategies' for the `NotificationService` to utilize, independently from each other. Then, utilizing the observer design pattern to 'publish' these notification triggers, and handle them in `NotificationService`'s `handleNotificationEvent()` method--through Java's [Spring Events](https://www.baeldung.com/spring-events).

Previously, every service class that sends notifications would both have to:
1. Customize that notification for its own purpose (i.e. including the event details, chat message details, etc.)
2. They'd have to construct & send the notification directly through Apple's Push Notification Service.

Now, 
1. Each service won't have to deal with customizing the notification, part just sending in parameters that are useful for that notification's specific 'strategy'. 
2. Each service class shouldn't have to deal with that second part of constructing & sending the notification directly, as we have a new `NotificationService` that interfaces with apple's Push Notification Service (+ possibly Firebase for Android in the future). 

Thus, we're decoupling all of this logic into 3 separate parts, instead of coupling it all into the originating service class by:
1. Service class triggers the notification, and instantiates that notification strategy
2. That notification strategy gets handled in a separate class, which assembles the notification simply using properties passed in from the service class
3. That notification, once constructed, is sent by the new `NotificationService`

## Usage:

- Create a new event in `spawn/Events/`

image

- Publish this event (like in the pub-sub pattern) as such:

image

- This will instantiate the event and pass it into the `ApplicationEventPublisher`, which seems to be Java's way of implementing the mediator design pattern:

image

image

- This will then be sent via the `NotificationService`'s `handleNotificationEvent` method that listens to the published event triggered in the service class:

image


ShaneMander (approved):
Nice, I like the application of the patterns a lot. Clean

PR #284, opened 2025-03-25
Refactor Service Classes

ShaneMander:
Breaking apart some of the service classes and introducing smaller more specific ones:

- IEventUserService

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/EventController.java
@@ -8,6 +8,7 @@ import com.danielagapov.spawn.Enums.ParticipationStatus;
 import com.danielagapov.spawn.Exceptions.Base.BaseNotFoundException;
 import com.danielagapov.spawn.Exceptions.EventsNotFoundException;
 import com.danielagapov.spawn.Services.Event.IEventService;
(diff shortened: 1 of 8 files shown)


Daggerpov:
I think what you have is great! It's clear why we need to separate methods that are related to different parts (basic CRUD, participation, etc.) into different classes.

I think splitting it up entirely makes the code a little harder to trace, though, and would personally prefer to have a hierarchy between the classes, where EventService is essentially the highest-up class that the controllers communicate with (as we had before), that then calls down to EventUserService through a Facade pattern, instead of completely separating the two.

So, what you have is 90% what I'd like, and I'm of course willing to talk through this solution, but if we were to go with my idea, that'd simply involve adding the methods back to EventService, and having them call down to EventUserService, like so:

```
private List<UUID> getParticipatingUserIdsByEventId(UUID eventId){
    return eventUserService.getParticipatingUserIdsByEventId(eventId);
}
```

Instead of the controllers directly calling different sub-service methods, because that can be trickier to track.

I think going with this approach, we can keep the simplicity we had before of the controllers and external classes calling the same methods, but internally, we're implementing those methods' logic in a 'deeper' service class layer through the Facade pattern, to keep the external usage the same and avoid all of the external refactoring that's coming with these changes.

PR #285, opened 2025-03-28
Notifications: Preference-Based Sending + Bug Fixes

Daggerpov:
- Only sending notifications if user has turned on that preference through their device
- Adjusted chat message sending according to Shane on Slack:
"Backend noti bug: I got a notification from commenting on my own Event when I should receive 0"


ShaneMander (approved):
Everything seems to look fine, but could you point out to me the flaw with the original logic that was causing the bug for a notification on a user's own comment? I couldn't actually find the issue with the original logic.

Issue #249, opened 2025-03-06
Delete `FullUserDTO` + Change OAuth endpoints to `BaseUserDTO`


Daggerpov:
Return `BaseUserDTO `from OAuth's endpoints: `make-user` & `sign-in`
<img width="285" alt="Image" src="https://github.com/user-attachments/assets/c60e92a4-72ea-44b6-85a2-c3565de489f4" />

Daggerpov:
Doing in current PR in review; closing

PR #288, opened 2025-03-28
`FullUserDTO` extinction; replaced all with `BaseUserDTO`

Daggerpov:
#249 

As `FullUserDTO` isn't being used anywhere in the front-end


src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -83,11 +86,22 @@ public ResponseEntity<List<? extends AbstractUserDTO>> getUserFriends(@PathVaria
...
+                List<UserDTO> users = userService.getUsersByTagId(tagId);
+                List<Map<String, Object>> enrichedUsers = new ArrayList<>();
+                
+                for (UserDTO user : users) {
+                    Map<String, Object> enrichedUser = new HashMap<>();
+                    enrichedUser.put("user", user);
+                    enrichedUser.put("friends", userService.getFriendsByUserId(user.getId()));
+                    enrichedUser.put("friendTags", friendTagService.getFriendTagsByOwnerId(user.getId()));
+                    enrichedUsers.add(enrichedUser);
+                }
```
ShaneMander, line 102: Should probably be service method, no?

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -20,7 +19,7 @@ public interface IUserService {
     // CRUD operations:
     UserDTO getUserById(UUID id);
 
-    FullUserDTO getFullUserById(UUID id);
+    BaseUserDTO getFullUserById(UUID id);
```
ShaneMander, line 22: Can remove this because we already have `IUserService::getBaseUserById`

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -32,9 +31,9 @@ public interface IUserService {
 
     UserDTO saveUserWithProfilePicture(UserDTO user, byte[] profilePicture);
 
-    FullUserDTO getFullUserByEmail(String email);
+    BaseUserDTO getFullUserByEmail(String email);
```
ShaneMander, line 34: Can be removed

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -32,9 +31,9 @@ public interface IUserService {
...
 
-    FullUserDTO getFullUserByUserEntity(User user);
+    BaseUserDTO getFullUserByUserEntity(User user);
```
ShaneMander, line 36: Can be removed

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -77,13 +76,13 @@ public interface IUserService {
     List<UUID> getInvitedUserIdsByEventId(UUID eventId);
 
     // Helper
-    FullUserDTO getFullUserByUser(UserDTO user, Set<UUID> visitedUsers);
+    BaseUserDTO getFullUserByUser(UserDTO user, Set<UUID> visitedUsers);
```
ShaneMander, line 79: This method can be removed and each usage of this method can be replaced with calls to `UserMapper::toBaseDTO` instead

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -77,13 +76,13 @@ public interface IUserService {
...
 
-    List<FullUserDTO> convertUsersToFullUsers(List<UserDTO> users, Set<UUID> visitedUsers);
+    List<BaseUserDTO> convertUsersToFullUsers(List<UserDTO> users, Set<UUID> visitedUsers);
```
ShaneMander, line 81: Can be removed and can replace old calls to this method with a new `UserMapper::toBaseDTOList` method

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
```diff
@@ -77,13 +76,13 @@ public interface IUserService {
...
 
     boolean existsByEmail(String email);
 
-    FullUserDTO getFullUserByUsername(String username);
+    BaseUserDTO getFullUserByUsername(String username);
```
ShaneMander, line 85: Can be removed since you've deleted the only usage

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
ShaneMander: Or replace it with a call to the new `UserService::getUserWithFriendsAndTags`

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -7,28 +7,31 @@
...
-import java.util.UUID;
+import java.util.*;
 
 @RestController()
 @RequestMapping("api/v1/users")
 public class UserController {
     private final IUserService userService;
     private final IS3Service s3Service;
     private final ILogger logger;
+    private final IFriendTagService friendTagService;
```
ShaneMander, line 27: Can be removed when you replace the logic in `UserController::getUsersByFriendTag` with the call to `UserService::getUserWithFriendsAndTags`

src/main/java/com/danielagapov/spawn/DTOs/DeviceTokenDTO.java
```diff
@@ -22,4 +22,9 @@ public boolean equals(Object obj) {
         DeviceTokenDTO that = (DeviceTokenDTO) obj; // Safe cast
         return token != null && token.equals(that.token); // Compare tokens
     }
+    
+    @Override
+    public int hashCode() {
```
ShaneMander, line 27: I might've missed it but where is this being used?

ShaneMander (changes requested):
Nice stuff, I've just left some comments on methods that can be removed since we no longer have a FullUserDTO

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
Daggerpov: Yea, service method

src/main/java/com/danielagapov/spawn/Services/User/IUserService.java
Daggerpov: Oh nice. Good catch!

src/main/java/com/danielagapov/spawn/DTOs/DeviceTokenDTO.java
Daggerpov: I think I was getting warnings that classes are supposed to have either none of those two methods, equals() or hashCode() or both.

Daggerpov: `java: Not generating hashCode: One of equals or hashCode exists. You should either write both of these or none of these (in the latter case, lombok generates them).`

ShaneMander (approved):

src/main/java/com/danielagapov/spawn/Controllers/UserController.java
```diff
@@ -47,25 +50,6 @@ public ResponseEntity<List<? extends AbstractUserDTO>> getUsers(@RequestParam(va
         }
     }
 
-    // TL;DR: Don't remove this endpoint; it may become useful.
-    @Deprecated(since = "Not being used on mobile currently.")
-    // full path: /api/v1/users/{id}?full=full
-    @GetMapping("{id}")
-    public ResponseEntity<AbstractUserDTO> getUser(@PathVariable UUID id, @RequestParam(value = "full", required = false) boolean full) {
```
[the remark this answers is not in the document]
Daggerpov: Yea, it was just messy with the deprecated `full` stuff; if we need it, we'll do it again, just as you're saying 👍

PR #289, opened 2025-03-29
Feedback Image Support

Daggerpov:
Made the endpoint the same as with the `make-user` one in `AuthController.java`, accepting the image data as a byte array. 

We discussed on Slack that feedback should take in an image, so I'll make the UI on mobile to go with this endpoint.


ShaneMander (approved):
Nice

PR #297, opened 2025-03-31
User Blocking

### LINKED ISSUE #173: (App Store Compliance) User Block Another User
A user needs to be able to block another user. I’m thinking this should result in:
- removing each other as friends (which we already have a UserService method for)
- them not showing up in each others’ recommended friends + incoming friend requests
- them not showing up in each others’ search friends results 

I’m thinking we have an entity that’s a 1-to-1 between a user to another user called, BlockedUser

evannawfal:
This PR adds support for blocking users on Spawn

### Key Changes
- Created BlockedUser Model, controller, service class, DTO, BlockedUserServiceTests, and mapper
- When a User is blocked, we now:
  - Remove the users as friends
  - Remove friend requests from the blocking user to the blocked user
  - Blocked users won’t appear in each other’s:
    - recommended friends
    - incoming friend requests
    - friend search results

Closes #173

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
@@ -35,10 +35,10 @@ public class BlockedUserController {
         }
     }
 
-    // DELETE /api/v1/blocked-users/{blockerId}/unblock/{blockedId}
-    @DeleteMapping("{blockerId}/unblock/{blockedId}")
-    public ResponseEntity<Void> unblockUser(@PathVariable UUID blockerId,
-                                            @PathVariable UUID blockedId) {
+    // DELETE /api/v1/blocked-users/unblock?blockerId=...&blockedId=...
+    @DeleteMapping("unblock")
+    public ResponseEntity<Void> unblockUser(@RequestParam UUID blockerId,
+                                            @RequestParam UUID blockedId) {
         try {
             blockedUserService.unblockUser(blockerId, blockedId);
             return new ResponseEntity<>(HttpStatus.NO_CONTENT);
--- a/src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
+++ b/src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
@@ -1,6 +1,7 @@
 package com.danielagapov.spawn.Controllers;
 
-import com.danielagapov.spawn.DTOs.BlockedUserDTO;
+import com.danielagapov.spawn.DTOs.BlockedUser.BlockedUserCreationDTO;
+import com.danielagapov.spawn.DTOs.BlockedUser.BlockedUserDTO;
 import com.danielagapov.spawn.Services.BlockedUser.IBlockedUserService;
 import com.danielagapov.spawn.Services.FriendRequest.IFriendRequestService;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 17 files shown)


src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
```diff
@@ -0,0 +1,101 @@
...
+    public BlockedUserController(IBlockedUserService blockedUserService, IFriendRequestService friendRequestService) {
+        this.blockedUserService = blockedUserService;
+        this.friendRequestService = friendRequestService;
+    }
+
+    // POST /api/v1/blocked-users/{blockerId}/block/{blockedId}
+    @PostMapping("{blockerId}/block/{blockedId}")
+    public ResponseEntity<Void> blockUser(@PathVariable UUID blockerId,
+                                          @PathVariable UUID blockedId,
+                                          @RequestParam(required = false) String reason) {
```
Daggerpov, line 29: Hey, you could rewrite this for a simpler endpoint by using 2 request params, or a DTO, not two path variables
```suggestion
    // POST /api/v1/blocked-users/block
    @PostMapping("block")
    public ResponseEntity<Void> blockUser(@RequestBody BlockUserCreationDTO) {
```

src/main/java/com/danielagapov/spawn/Repositories/IBlockedUserRepository.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+
+@Repository
+public interface IBlockedUserRepository extends JpaRepository<BlockedUser, UUID> {
+
+    boolean existsByBlocker_IdAndBlocked_Id(UUID blockerId, UUID blockedId);
+    Optional<BlockedUser> findByBlocker_IdAndBlocked_Id(UUID blockerId, UUID blockedId);
```
Daggerpov, line 15: Great use of optional here, for our new code practices 👍

src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
```diff
@@ -0,0 +1,101 @@
...
+        if (blockerId == null || blockedId == null) {
```
Daggerpov, line 30: After change above to using a new DTO, this won't be needed, and this will actually automatically be done through Java Spring if you don't mark these attributes as optional, **I'm pretty sure**--as that seems to be the case with other DTOs; I get 400 error codes on mobile when I forget to include a property

src/main/java/com/danielagapov/spawn/Services/FriendRequest/IFriendRequestService.java
```diff
@@ -20,4 +20,5 @@ public interface IFriendRequestService {
     List<FetchFriendRequestDTO> convertFriendRequestsToFetchFriendRequests(List<CreateFriendRequestDTO> friendRequests);
 
     List<CreateFriendRequestDTO> getSentFriendRequestsByUserId(UUID userId);
+    void deleteFriendRequestBetweenUsers(UUID senderId, UUID receiverId);
```
Daggerpov, line 23:
```suggestion
    void deleteFriendRequestBetweenUsersIfExists(UUID senderId, UUID receiverId);
```
For clarity

src/main/java/com/danielagapov/spawn/Services/FriendRequest/FriendRequestService.java
```diff
@@ -126,6 +133,19 @@ public void deleteFriendRequest(UUID id) {
         }
     }
 
+    @Override
+    public void deleteFriendRequestBetweenUsers(UUID senderId, UUID receiverId) {
+        try {
+            List<FriendRequest> requests = repository.findBySenderIdAndReceiverId(senderId, receiverId);
```
Daggerpov, line 139: I agree that we shouldn't throw an exception if we can't find any requests, but let's change the method name as I said above to include `IfExists` for clarity that this is the internal behaviour, and that we're not throwing an error if there isn't a friend req in either direction

src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
```diff
@@ -0,0 +1,101 @@
...
+            blockedUserService.blockUser(blockerId, blockedId, reason);
+
+            return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+        } catch (Exception e) {
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+
+    // DELETE /api/v1/blocked-users/{blockerId}/unblock/{blockedId}
```
Daggerpov, line 46: same idea as with the other endpoint, to use a DTO or request params instead of multiple path vars

src/main/java/com/danielagapov/spawn/Controllers/BlockedUserController.java
```diff
@@ -0,0 +1,101 @@
...
+            return ResponseEntity.ok(blockedUserService.isBlocked(blockerId, blockedId));
+        } catch (Exception e) {
+            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+
+    // GET /api/v1/blocked-users/{blockerId}
+    @GetMapping("{blockerId}")
+    public ResponseEntity<List<BlockedUserDTO>> getBlockedUsers(@PathVariable UUID blockerId) {
```
Daggerpov, line 74: Let's combine this endpoint and `getBlockedUserIds` into one, with a requestparam called `ids` that is non required, and will toggle between returning the users or ids.

Daggerpov: For simplicity, and generally, the less endpoints the better (for maintaining the code and debugging)

Daggerpov (changes requested):
Great work! I like how you implemented this from the top-down! Left some minor comments

Daggerpov:
Merging into staging because it'll affect mobile as it stands rn, so I put those PRs into staging to ensure mobile is ready

src/main/java/com/danielagapov/spawn/Repositories/IBlockedUserRepository.java
evannawfal: B)

Daggerpov (approved):

PR #305, opened 2025-04-05
Change getRecommendedFriendsBySearch

ShaneMander:
Small PR for the existing search endpoint to include results retrieved from the new fuzzy search implementation.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Models/User.java
+++ b/src/main/java/com/danielagapov/spawn/Models/User.java
@@ -20,6 +20,13 @@ import java.util.UUID;
 @AllArgsConstructor
 @Getter
 @Setter
+@Table(
+        name = "user",
+        indexes = {
+                @Index(name = "idx_first_name", columnList = "first_name"),
+                @Index(name = "idx_last_name", columnList = "last_name")
+        }
+)
 public class User implements Serializable {
     @Id
     @GeneratedValue
@@ -32,6 +39,8 @@ public class User implements Serializable {
     private String firstName;
     private String lastName;
     private String bio;
+    
+    @Column(nullable = true, unique = true)
     private String email;
     private String password;
     private boolean verified;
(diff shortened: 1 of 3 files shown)


src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -20,6 +20,13 @@
 @AllArgsConstructor
 @Getter
 @Setter
+@Table(
+        name = "user",
+        indexes = {
+                @Index(name = "idx_first_name", columnList = "first_name"),
```
Daggerpov, line 26: These seem like smart indexes

src/main/java/com/danielagapov/spawn/Models/User.java
```diff
@@ -32,6 +39,8 @@ public class User implements Serializable {
     private String firstName;
     private String lastName;
     private String bio;
+    
+    @Column(nullable = true, unique = true)
```
Daggerpov, line 43: How come this needs to be nullable?

src/main/java/com/danielagapov/spawn/Services/UserSearch/UserSearchService.java
```diff
@@ -210,6 +219,12 @@ private Set<UUID> getExcludedUserIds(UUID userId) {
 
     @Override
     public List<BaseUserDTO> searchByQuery(String searchQuery) {
+        List<User> users = searchUsersByQuery(searchQuery);
```
Daggerpov, line 222: Hey, could we remove this method and `return UserMapper.toDTOList(users)` straight from `searchUsersByQuery()` for simplicity?

Daggerpov (approved):
Left an optional improvement. Good work!

src/main/java/com/danielagapov/spawn/Models/User.java
Daggerpov: Ah ok

PR #306, opened 2025-04-09
`LoggingUtils` for better user info logging (not just user id)

Daggerpov:
This is to address logs being hard to read, when they involve users, being that we've historically only logged the user id. 

- Also refactored `NotificationService`
- Also Refactored user deletion method to void instead of boolean return type

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/Auth/AuthService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/Auth/AuthService.java
@@ -42,7 +42,7 @@ public class AuthService implements IAuthService {
         checkIfUniqueCredentials(authUserDTO);
         try {
             UserDTO userDTO = createAndSaveUser(authUserDTO);
-            User user = userService.getUserEntityById(userDTO.getId());
+            User user = UserMapper.toEntity(userDTO);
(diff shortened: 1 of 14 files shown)


src/main/java/com/danielagapov/spawn/Services/Auth/AuthService.java
```diff
@@ -37,57 +38,72 @@ public class AuthService implements IAuthService {
 
     @Override
     public UserDTO registerUser(AuthUserDTO authUserDTO) throws FieldAlreadyExistsException {
+        logger.info("Attempting to register new user with username: " + authUserDTO.getUsername());
         checkIfUniqueCredentials(authUserDTO);
         try {
             UserDTO userDTO = createAndSaveUser(authUserDTO);
+            User user = userService.getUserEntityById(userDTO.getId());
```
ShaneMander, line 45: Can we use a mapper here instead?

ShaneMander (changes requested):
Nice, just left 1 small comment

evannawfal (approved):

src/main/java/com/danielagapov/spawn/Services/Auth/AuthService.java
Daggerpov: Addressed

PR #308, opened 2025-04-12
TTL Implemented

evannawfal:
Added Time to live (TTL) to remove stale cached data. The current time until data expires is 10 minutes, let me know if you have any thoughts on this. Also started to try and fix the cause of that friendtag caching bug.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
+++ b/src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
@@ -23,6 +23,8 @@ import com.danielagapov.spawn.Models.User;
 import com.danielagapov.spawn.Repositories.*;
 import com.danielagapov.spawn.Services.FriendTag.IFriendTagService;
 import com.danielagapov.spawn.Services.User.IUserService;
+import org.springframework.cache.annotation.CacheEvict;
+import org.springframework.cache.annotation.Caching;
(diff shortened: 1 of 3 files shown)


Daggerpov:
I feel like if we think of app usage, TTL should be much higher, like even a day or so. 

If I understand correctly, if the data changes, it'd get invalidated automatically, right, @evannawfal ? So, since this TTL is just for the maximum time until it gets expired, I don't think we need such a strict time. I could definitely be misguided, but that's just my opinion, and through talks with ChatGPT. 

image

Daggerpov:
I thought I already fixed the friend tag caching bug, no? @evannawfal

src/main/java/com/danielagapov/spawn/Services/ChatMessage/ChatMessageService.java
```diff
@@ -111,6 +113,12 @@ public ChatMessageDTO getChatMessageById(UUID id) {
     }
 
     @Override
+    @Caching(evict = {
+            @CacheEvict(value = "eventById", key = "#newChatMessageDTO.eventId"),
+            @CacheEvict(value = "fullEventById", allEntries = true),
```
Daggerpov, line 118: Spot on!
Seems like you referenced which of our event endpoints are in use; great work!

Daggerpov (approved):
Looks great! Left some thoughts, but the code is all good, and you could adjust the TTL later

PR #312, opened 2025-04-19
Cache domain

Daggerpov:
The main logic and cached types can be found in `CacheService`

TODO:
- [ ] Run SQL Scripts once merged for new `lastUpdated` fields
- [x] Address Shane's comment about unimplemented methods

Matching with mobile's [Caching Groundwork #169](https://github.com/Daggerpov/Spawn-App-iOS-SwiftUI/pull/169)


README.md
```diff
@@ -67,6 +68,134 @@ Spring Boot is our back-end framework with the Java language. It handles the API
...
+  }
+}
+```
+
+#### Push Notification Handling
+
+The app listens for push notifications with specific types that indicate data changes:
+
+- `friend-accepted`: When a friend request is accepted
+- `event-updated`: When an event is updated
```
ShaneMander, line 143: Should we also create some notification for when a user gets blocked? Since that would affect the friends cache?

ShaneMander: Event creation as well?

ShaneMander (commented):
Is this PR ready to be merged with the missing implementation of `getLatestEventActivity` and `getLatestFriendActivity`? 

Also.. 
This caching behavior is for just the user's friends, and the events they are invited to or have created? 

When the client has a validated cache, which API calls are no longer made from the client?

README.md
Daggerpov: Yup, those both make sense

src/main/java/com/danielagapov/spawn/Services/FriendRequest/FriendRequestService.java
```diff
@@ -252,4 +253,22 @@ public List<CreateFriendRequestDTO> getSentFriendRequestsByUserId(UUID userId) {
             throw e;
         }
     }
+
+    @Override
+    public Instant getLatestFriendRequestTimestamp(UUID userId) {
```
ShaneMander, line 258: Not implemented

ShaneMander (changes requested):
Looks like one method (FriendRequestService::getLatestFriendRequestTimestamp) didn't get implemented

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -334,13 +335,11 @@ private EventDTO constructDTOFromEntity(Event eventEntity) {
 
     @Override
     @Caching(evict = {
-            @CacheEvict(value = "eventById", key = "#id"),
+            @CacheEvict(value = "eventById", key = "#result.id"),
```
evannawfal, line 338: I think since the method returns a boolean, using `#result.id` in the annotation will cause an error because result refers to the return value, and you can't access `.id` on a boolean

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
```diff
@@ -334,13 +335,11 @@ private EventDTO constructDTOFromEntity(Event eventEntity) {
...
             @CacheEvict(value = "fullEventById", allEntries = true),
-            @CacheEvict(value = "eventsByOwnerId", allEntries = true),
+            @CacheEvict(value = "eventsByOwnerId", key = "#result.creatorUserId"),
```
evannawfal, line 340: Same here with result

evannawfal (changes requested):

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Pushed this change. Is this the right approach, @evannawfal ? 
image

ShaneMander (approved):
I approve if @evannawfal's comments are also resolved

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Changed back. Resolving

src/main/java/com/danielagapov/spawn/Services/Event/EventService.java
Daggerpov: Changed back. Resolving

PR #331, opened 2025-05-05
Redis Caching and Eviction for Event Feeds, Friends, Recommendations, and Friend Tags

evannawfal:
This PR completes and corrects the current Redis caching implementation:

- **Event Feeds**
- **Friends**
- **Recommended Friends**
- **Friend Tags**

Some evictions are handled manually inside functions (via `cacheManager`) rather than through `@CacheEvict` annotations. This was necessary in cases where the required cache key (e.g. `ownerId`) was not available in the method parameters or return types. 

If you have ideas for cleaner alternatives without significant refactoring please let me know.


Daggerpov (approved):

ShaneMander (approved):
I think there is a use case for the Strategy pattern here for evictions made using the `cacheManager`, similar to how we use `NotificationEvent`, however, we could make that refactor in a future PR.

PR #334, opened 2025-05-06
Change Password

Daggerpov:


src/main/java/com/danielagapov/spawn/Controllers/User/AuthController.java
```diff
@@ -178,6 +178,44 @@ public ModelAndView verifyEmail(@RequestParam("token") String emailToken) {
...
+    @PostMapping("change-password")
+    public ResponseEntity<?> changePassword(@RequestBody PasswordChangeDTO passwordChangeDTO, HttpServletRequest request) {
+        try {
+            logger.info("Password change request received");
+            
+            // Extract username from JWT token
+            final String authHeader = request.getHeader("Authorization");
+            if (authHeader == null || !authHeader.startsWith("Bearer ")) {
+                return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(new ErrorResponse("Authentication required"));
+            }
```
ShaneMander, line 191: Can remove since this is checked in the filter chain

src/main/java/com/danielagapov/spawn/Controllers/User/AuthController.java
```diff
@@ -178,6 +178,44 @@ public ModelAndView verifyEmail(@RequestParam("token") String emailToken) {
...
+            
+            final String token = authHeader.substring(7);
+            final String username = jwtService.extractUsername(token);
+            
+            if (username == null) {
+                return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(new ErrorResponse("Invalid token"));
+            }
```
ShaneMander, line 198: Can remove since this is checked in the filter chain

ShaneMander (changes requested):
Looks good. Left comments on parts that can be removed because it is repeated from the filter chain, and so removing them from the `changePassword` endpoint would be more concise.

PR #344, opened 2025-05-13
Notify Team about Beta Sign Ups + Clean Logs

Daggerpov:
- Kept logs for features in development or for ones that haven't been tested yet, like email/pass user registration, change password, etc.
    - Cleaned up ones that were clogging the Railway logs, though, like for tokens. 
- Notify team by email upon beta access sign up, since we haven't sent a follow-up invite since March but should, since new sign ups have been made
- Move logic of getting DTO properties from controller -> service layer for `UserService::updateUser()`


ShaneMander (approved):
Looks good to me

PR #352, opened 2025-05-18
Google User Id -> Token

Daggerpov:


src/main/java/com/danielagapov/spawn/Controllers/User/AuthController.java
```diff
@@ -76,13 +93,29 @@ public ResponseEntity<?> signIn(@RequestParam("externalUserId") String externalU
...
         try {
-            BaseUserDTO user = oauthService.createUser(userCreationDTO, externalUserId, provider);
+            BaseUserDTO user;
+            
+            // Handle Google ID token authentication if present
+            if (idToken != null && !idToken.isEmpty()) {
+                user = oauthService.createUserWithGoogleToken(userCreationDTO, idToken);
+            } 
+            // Fall back to externalUserId for Apple or backward compatibility
+            else if (externalUserId != null && !externalUserId.isEmpty() && provider != null) {
```
ShaneMander, line 109: We should change this condition to be `else if (... && provider == APPLE)`, otherwise we still allow google login without tokens

ShaneMander (changes requested):

src/main/java/com/danielagapov/spawn/Controllers/User/AuthController.java
Daggerpov: Makes sense!

ShaneMander:
I think we should also refactor apple oauth to use id tokens there as well

ShaneMander (approved):
Looks good

PR #368, opened 2025-06-25
Activity type pinning

Daggerpov:
- Adjust endpoints to `/api/v1/{userId}/activity-types` to follow REST hierarchical resource scheme
- Add ability to pin an activity type through endpoint and persisting in back-end
- Add `ownerUserId` to `ActivityTypeDTO`
- Add service class tests
- Fixed config errors with the repo tests:
image


ShaneMander (changes requested):
Left some questions about how mobile is supposed to interact with these endpoints + some reorganizing suggestions and a few other comments.

I don't think I see the benefit in adding the `ownerUserId` field to the ActivityTypeDTO. I think there is a greater redundancy in adding that as a field than having it as a separate argument to some of other service methods. Let's continue this discussion in Slack.

src/main/java/com/danielagapov/spawn/Controllers/ActivityTypeController.java
```diff
@@ -4,39 +4,147 @@
...
+        this.activityTypeService = activityTypeService;
+        this.logger = logger;
+    }
+
+    /**
+     * Get activity types owned by a user
+     * GET /api/v1/{userId}/activity-types
+     */
+    @GetMapping
+    public ResponseEntity<List<ActivityTypeDTO>> getOwnedActivityTypesForUser(@PathVariable UUID userId) {
```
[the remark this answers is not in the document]
Daggerpov: They do the same thing. The old one also called `return ResponseEntity.ok(activityTypeService.getActivityTypesByUserId(userId));`, so yea, I guess Cursor just reordered the endpoints and re-wrote the endpoint comment.

src/main/java/com/danielagapov/spawn/Controllers/ActivityTypeController.java
```diff
@@ -4,39 +4,147 @@
...
     }
 
-    @PostMapping("{userId}")
-    public ResponseEntity<?> updateActivityTypes(@PathVariable UUID userId, @RequestBody BatchActivityTypeUpdateDTO batchUpdateDTO) {
+    /**
+     * Delete an activity type
+     * DELETE /api/v1/{userId}/activity-types/{activityTypeId}
+     */
+    @DeleteMapping("/{activityTypeId}")
+    public ResponseEntity<Void> deleteActivityType(
```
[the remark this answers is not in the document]
Daggerpov: Diddo as above; forgot about the plan from a month ago

src/main/java/com/danielagapov/spawn/DTOs/ActivityType/ActivityTypeDTO.java
```diff
@@ -16,4 +16,6 @@ public class ActivityTypeDTO implements Serializable {
     private List<BaseUserDTO> associatedFriends;
     private String icon;
     private int orderNum;
+    private UUID ownerUserId;
+    private Boolean isPinned;
```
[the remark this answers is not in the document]
Daggerpov: Yea, this is something I was talking to Ethan about implementing, and it's now done in Figma. Hopefully that illustrates this:

image

src/main/java/com/danielagapov/spawn/Services/ActivityType/IActivityTypeService.java
```diff
@@ -9,12 +9,51 @@
...
+     */
+    void deleteActivityType(UUID activityTypeId);
+    
+    /**
+     * Toggle the pinned status of an activity type
+     * @param activityTypeId The activity type ID to toggle pin status
+     * @param userId The user ID who owns the activity type (for authorization)
+     * @return The updated activity type DTO
+     */
+    ActivityTypeDTO togglePin(UUID activityTypeId, UUID userId);
```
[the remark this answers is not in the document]
Daggerpov: Yea, I think you're right that we could incorporate it into the batch update, since it's a property on `ActivityTypeDTO` now

src/main/java/com/danielagapov/spawn/Controllers/ActivityTypeController.java
```diff
@@ -4,39 +4,147 @@
...
+            // Create a new DTO with the ownerUserId set from the path parameter
+            ActivityTypeDTO dtoWithOwner = new ActivityTypeDTO(
+                activityTypeDTO.getId(),
+                activityTypeDTO.getTitle(),
+                activityTypeDTO.getAssociatedFriends(),
+                activityTypeDTO.getIcon(),
+                activityTypeDTO.getOrderNum(),
+                userId, // Set ownerUserId from path parameter
+                activityTypeDTO.getIsPinned() != null ? activityTypeDTO.getIsPinned() : false // Handle isPinned
+            );
```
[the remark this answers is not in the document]
Daggerpov: Interesting suggestion from Claude
image

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
```diff
@@ -23,24 +25,65 @@ public class ActivityTypeService implements IActivityTypeService {
...
+        try {
+            if (activityTypeDTOs.getUpdatedActivityTypes().isEmpty() && activityTypeDTOs.getDeletedActivityTypeIds().isEmpty()) {
+                throw new IllegalArgumentException("No activity types to update or delete");
+            }
+
+            // Set ownerUserId for all updated activity types using mapper
+            List<ActivityTypeDTO> updatedWithOwner = ActivityTypeMapper.withOwnerUserId(
+                activityTypeDTOs.getUpdatedActivityTypes(), 
+                userId
+            );
```
ShaneMander, line 75: I think this step is redundant. We should just be converting straight to ActivityType entities. I don't see the need for an intermediate DTO

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
```diff
@@ -23,24 +25,65 @@ public class ActivityTypeService implements IActivityTypeService {
...
     public List<ActivityTypeDTO> getActivityTypesByUserId(UUID userId) {
+        // Returns activity types owned by this user
         return ActivityTypeMapper.toDTOList(repository.findActivityTypesByCreatorId(userId));
     }
 
+
+
     @Override
-    public void updateActivityTypes(UUID userId, BatchActivityTypeUpdateDTO activityTypeDTOs) {
+    public BatchActivityTypeUpdateDTO updateActivityTypes(BatchActivityTypeUpdateDTO activityTypeDTOs) {
```
ShaneMander, line 35: Related to my other comment, I think we can squash this method into the other `updateActivityTypes` that takes arguments `userId` and `BatchActivityTypeUpdateDTO`. We aren't meaningfully using the fact that the `ownerUserId` field is set in each of the activity types. We only retrieve it once on line 49 but that value can also just be retrieved from the method parameter if it was one combined method.

src/main/java/com/danielagapov/spawn/Services/ActivityType/IActivityTypeService.java
```diff
@@ -9,12 +9,38 @@
...
 
-    void updateActivityTypes(UUID userId, BatchActivityTypeUpdateDTO activityTypeDTOs);
-
+    /**
+     * Batch update activity types (create, update, delete) with userId
+     * @param userId The user ID who owns the activity types
+     * @param batchActivityTypeUpdateDTO The batch update data (activity types do not need ownerUserId)
+     * @return The batch update result
+     */
+    BatchActivityTypeUpdateDTO updateActivityTypes(UUID userId, BatchActivityTypeUpdateDTO batchActivityTypeUpdateDTO);
```
ShaneMander, line 33: If we intend on returning a list of updated activity type dtos back to the client, then we probably should return `List<ActivityTypeDTO>` and not a `BatchActivityTypeUpdateDTO`

src/main/java/com/danielagapov/spawn/Models/ActivityType.java
```diff
@@ -34,11 +33,27 @@ public class ActivityType {
     private Integer orderNum;
```
ShaneMander, line 33: I just realized that while it is true we want the `orderNum` to be unique (for a given user's activity types), adding the Unique Constraint makes updates difficult since to swap the orderNum of two activity types, it would mean having to first save the orderNum value of one activity type to some temporarily value then performing the swap, and then saving, and I imagine this would be more complex with more activity types. 

I think it would be best to actually drop the unique constraint so that we can temporarily violate our implied unique constraint during updates, and instead add some careful application level logic so that we before/after updates, the orderNum is always unique across a user's given activity types.

I initially added this in my first pass of activity types so I don't mind working on this, but I just wanted to write it somewhere so I don't forget. If you wanted to tackle it in this PR as well though, that would be great, but up to you

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
```diff
@@ -23,24 +25,65 @@ public class ActivityTypeService implements IActivityTypeService {
...
+            
+            // Handle updates/creations - need userId for these operations
             if (!activityTypeDTOs.getUpdatedActivityTypes().isEmpty()) {
+                UUID userId = activityTypeDTOs.getUpdatedActivityTypes().get(0).getOwnerUserId();
+                User creator = userService.getUserEntityById(userId);
+                
                 logger.info("Saving updated or newly created activity types for user: " + creator.getUsername());
                 List<ActivityType> activityTypes = ActivityTypeMapper.toEntityList(activityTypeDTOs.getUpdatedActivityTypes(), creator);
                 repository.saveAll(activityTypes);
             }
```
ShaneMander, line 55: Just realized, should we be doing any sort of validation steps on the activity types to update? For example, making sure there are at most (or exactly?) X activity types that have `isPinned=true`. Making sure that the `orderNum` values are valid, i.e. in the range of [1, n] where n is the number of activity types owned by a given user.

ShaneMander (changes requested):
We should drop one of the `updateActivityTypes` methods and just leave one method that takes in a `userId` and `BatchActivityUpdateDTO`. Additionally, in our current implementation, we aren't making any meaningful use of the fact that an `ownerUserId` is attached to the `ActivityTypeDTO`. I also left a suggestion for some data validation before we update an activity type.

Additionally, I left a note about dropping the unique constraint from the `orderNum` property in the entity -- up to you if you want to do this now in this PR or I can tackle it in a separate one

src/main/java/com/danielagapov/spawn/Models/ActivityType.java
Daggerpov: This makes a lot of sense; I'll do it

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
Daggerpov: Makes sense; will do this

Daggerpov: Adding tests for this now: 
image

src/main/java/com/danielagapov/spawn/Models/ActivityType.java
Daggerpov: Done

src/main/java/com/danielagapov/spawn/Controllers/ActivityTypeController.java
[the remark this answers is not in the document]
Daggerpov: Done, I'm pretty sure. These are the mappings:

```
public static ActivityTypeDTO toDTO(ActivityType entity) {
        return new ActivityTypeDTO(
                entity.getId(),
                entity.getTitle(),
                entity.getAssociatedFriends() != null ? UserMapper.toDTOList(entity.getAssociatedFriends()) : Collections.emptyList(),
                entity.getIcon(),
                entity.getOrderNum(),
                entity.getCreator().getId(),
                entity.getIsPinned()
        );
    }
```

```
public static ActivityType toEntity(ActivityTypeDTO dto, User creator) {
        return new ActivityType(
                dto.getId(),
                dto.getTitle(),
                dto.getAssociatedFriends() != null ? UserMapper.toEntityList(dto.getAssociatedFriends()) : Collections.emptyList(),
                creator,
                dto.getOrderNum(),
                dto.getIcon(),
                dto.getIsPinned() != null ? dto.getIsPinned() : false
        );
    }
```

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
Daggerpov: Fixed

src/main/java/com/danielagapov/spawn/Services/ActivityType/ActivityTypeService.java
Daggerpov: Removed this redundant method

PR #378, opened 2025-07-06
SMS verification

ShaneMander:
Don't merge this yet

SMS account verification which uses a 3rd party: Twilio Verify API

As you can see from `TwilioService.java`, this is a very simple integration which automatically handles security and rate limiting. The code expiry is 10 mins. I think this is suitable for our purposes and does not be changed.

### DIFF
--- a/src/main/java/com/danielagapov/spawn/Exceptions/SMSVerificationException.java
+++ b/src/main/java/com/danielagapov/spawn/Exceptions/SMSVerificationException.java
@@ -1,0 +1,7 @@
+package com.danielagapov.spawn.Exceptions;
+
+public class SMSVerificationException extends RuntimeException {
+    public SMSVerificationException(String message) {
+        super(message);
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,12 @@
             <artifactId>firebase-admin</artifactId>
             <version>9.4.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.twilio.sdk</groupId>
+            <artifactId>twilio</artifactId>
+            <version>10.9.1</version>
+            <scope>compile</scope>
+        </dependency>
         <!-- Flyway for database migrations - DISABLED -->
         <!-- Using Hibernate DDL auto-update instead of Flyway migrations -->
         <!--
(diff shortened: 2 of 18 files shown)


src/main/java/com/danielagapov/spawn/Models/PhoneNumberVerification.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+@Entity
+@Getter
+@Setter
+public class PhoneNumberVerification {
```
Daggerpov, line 16: Don't need anymore, right?

src/main/java/com/danielagapov/spawn/Exceptions/SMSVerificationException.java
```diff
@@ -0,0 +1,7 @@
+package com.danielagapov.spawn.Exceptions;
+
+public class SMSVerificationException extends RuntimeException {
```
Daggerpov, line 3: Don't need anymore, right?

src/main/java/com/danielagapov/spawn/Config/TestConfig.java
```diff
@@ -73,5 +73,10 @@ public String getDefaultProfilePicture() {
         public void deleteObjectByURL(String urlString) {
             // No-op for test
         }
+
+        @Override
```
Daggerpov, line 77: What's this for?

src/main/java/com/danielagapov/spawn/Repositories/IPhoneNumberVerificationRepository.java
```diff
@@ -0,0 +1,13 @@
+package com.danielagapov.spawn.Repositories;
+
+import com.danielagapov.spawn.Models.PhoneNumberVerification;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.UUID;
+
+public interface IPhoneNumberVerificationRepository extends JpaRepository<PhoneNumberVerification, UUID> {
```
Daggerpov, line 8: Don't need anymore, right?

Daggerpov (changes requested):
Seems like some phone verification stuff is still here, and might not be needed. Feel free to resolve my comments if you think they're wrong, though, let's not go back-n-forth; go with what you think.

src/main/java/com/danielagapov/spawn/Models/PhoneNumberVerification.java
ShaneMander: yes good catches, no phone number stuff is needed any more

ShaneMander: Addressed

Daggerpov (dismissed):

PR #399, opened 2025-07-13
Fix Google OAuth - Incomplete Registration -> then retry

Daggerpov:


src/main/java/com/danielagapov/spawn/Services/OAuth/AppleOAuthStrategy.java
```diff
@@ -56,42 +59,76 @@ public String verifyIdToken(String idToken) {
...
-            // Verify the token
-            DecodedJWT verifiedJWT = verifier.verify(idToken);
-
-            // Extract the subject (user ID)
-            String userId = verifiedJWT.getSubject();
-            logger.info("Successfully verified Apple ID token and extracted user ID: " + userId);
-
-            return userId;
-
+            // Use retry helper for token verification
```
ShaneMander, line 62: Why we changing the token verification? This was working, no?

ShaneMander:
Added a couple fixes for apple account creation and refactored auth response dto

PR #415, opened 2025-08-10
Auth bug

ShaneMander:
Fixed a null pointer exception and added a onboarding cancellation endpoint


Daggerpov:
Discussed don't need auth cancellation since onboarding continuation 2nd case (per Slack)

ShaneMander:
Removed onboarding cancellation

Daggerpov (approved):

